PR #4, opened 2023-10-10
refactor child contracts

dorin-iancu:
- remove EthAddress type
- remove fees logic
- remove EthTransaction -> use Transaction everywhere. Also add additional fields to allow cross-chain SC calls.
- refactor to allow multi-token transfers (max 10 per call atm)
- refactor to allow SFT/NFT transfers

Multi-sig SC will be fixed in a future PR.


common/token-module/src/lib.rs
```diff
@@ -3,88 +3,21 @@
...
-
     #[only_owner]
     #[endpoint(addTokenToWhitelist)]
-    fn add_token_to_whitelist(
-        &self,
-        token_id: TokenIdentifier,
-        ticker: ManagedBuffer,
-        opt_default_price_per_gas_unit: OptionalValue<BigUint>,
-    ) {
+    fn add_token_to_whitelist(&self, token_id: TokenIdentifier, ticker: ManagedBuffer) {
```
CostinCarabas, line 12: Can you extract `ticker` from `token_id` so `token_ticker` storage would not be necessary?

common/transaction/src/lib.rs
```diff
@@ -3,64 +3,124 @@
...
+            EsdtTokenType::Meta => StolenFromFrameworkEsdtTokenType::Meta,
+            EsdtTokenType::Invalid => StolenFromFrameworkEsdtTokenType::Invalid,
+        }
+    }
+}
+
+#[derive(
+    TopDecode, TopEncode, NestedDecode, NestedEncode, TypeAbi, Debug, ManagedVecItem, Clone,
+)]
+pub struct StolenFromFrameworkEsdtTokenData<M: ManagedTypeApi> {
```
CostinCarabas, line 71: Can you rename `StolenFromFrameworkEsdtTokenData`?

esdt-safe/src/lib.rs
```diff
@@ -41,15 +42,10 @@ pub trait EsdtSafe:
         self.first_batch_id().set_if_empty(1);
```
CostinCarabas, line 42: I think `first_batch_id` should be renamed with `current_batch_id`

common/token-module/src/lib.rs
dorin-iancu: Actually, we likely don't need ticker at all anymore. Future PR.

common/transaction/src/lib.rs
dorin-iancu: It will be introduced in framework in the next version. Will be deleted then :)

esdt-safe/src/lib.rs
```diff
@@ -4,33 +4,34 @@
...
 #[multiversx_sc::contract]
 pub trait EsdtSafe:
-    fee_estimator_module::FeeEstimatorModule
-    + token_module::TokenModule
+    token_module::TokenModule
     + tx_batch_module::TxBatchModule
     + max_bridged_amount_module::MaxBridgedAmountModule
     + multiversx_sc_modules::pause::PauseModule
 {
     /// fee_estimator_contract_address - The address of a Price Aggregator contract,
```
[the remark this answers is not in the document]
dorin-iancu: Done.

esdt-safe/src/lib.rs
```diff
@@ -84,13 +80,20 @@ pub trait EsdtSafe:
                     // local burn role might be removed while tx is executed
                     // tokens will remain locked forever in that case
                     // otherwise, the whole batch would fail
-                    if self.is_local_role_set(&tx.token_identifier, &EsdtLocalRole::Burn) {
-                        self.burn_esdt_token(&tx.token_identifier, &tx.amount);
+                    for token in &tx.tokens {
+                        if self.is_local_role_set(&token.token_identifier, &EsdtLocalRole::Burn) {
+                            self.send().esdt_local_burn(
```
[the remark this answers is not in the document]
dorin-iancu: That's why it's checked first :)

esdt-safe/src/lib.rs
dorin-iancu: Eh, maybe, but not for now. Let's leave it as it is.

esdt-safe/src/lib.rs
```diff
@@ -229,12 +226,19 @@ pub trait EsdtSafe:
     fn get_refund_amounts(
         &self,
         address: ManagedAddress,
-    ) -> MultiValueEncoded<MultiValue2<TokenIdentifier, BigUint>> {
+    ) -> MultiValueEncoded<MultiValue3<TokenIdentifier, u64, BigUint>> {
         let mut refund_amounts = MultiValueEncoded::new();
         for token_id in self.token_whitelist().iter() {
```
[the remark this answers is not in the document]
dorin-iancu: No need, the call is done only once.

PR #10, opened 2023-10-12
Send tokens

dorin-iancu:
Send the same tokens to the user instead of burn & mint.

Also remove `token_ticker`, since it was unused.

### DIFF
--- a/multi-transfer-esdt/src/token_mapping.rs
+++ b/multi-transfer-esdt/src/token_mapping.rs
@@ -1,0 +1,22 @@
+multiversx_sc::imports!();
+
+#[multiversx_sc::module]
+pub trait TokenMappingModule {
+    #[only_owner]
+    #[endpoint(setSovToMxTokenId)]
+    fn set_sov_to_mx_token_id(&self, sov_token_id: TokenIdentifier, mx_token_id: TokenIdentifier) {
+        require!(
+            sov_token_id.is_valid_esdt_identifier() && mx_token_id.is_valid_esdt_identifier(),
+            "Invalid token IDs"
+        );
+
+        self.sovereign_to_multiversx_token_id(&sov_token_id)
+            .set(mx_token_id);
+    }
+
+    #[storage_mapper("sovToMxTokenId")]
+    fn sovereign_to_multiversx_token_id(
+        &self,
+        sov_token_id: &TokenIdentifier,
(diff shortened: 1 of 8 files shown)


multi-transfer-esdt/src/token_mapping.rs
```diff
@@ -0,0 +1,22 @@
...
+pub trait TokenMappingModule {
+    #[only_owner]
+    #[endpoint(setSovToMxTokenId)]
+    fn set_sov_to_mx_token_id(&self, sov_token_id: TokenIdentifier, mx_token_id: TokenIdentifier) {
+        require!(
+            sov_token_id.is_valid_esdt_identifier() && mx_token_id.is_valid_esdt_identifier(),
+            "Invalid token IDs"
+        );
+
+        self.sovereign_to_multiversx_token_id(&sov_token_id)
```
CostinCarabas, line 13: Do you use the same storage for `multivers_to_sovereign`? If so, it's hard to get  `sov_token_id` from `mx_token_id`

multi-transfer-esdt/src/transfer_tokens.rs
```diff
@@ -77,23 +78,35 @@ pub trait TransferTokensModule:
...
     ) -> PaymentsVec<Self::Api> {
+        let own_sc_address = self.blockchain().get_sc_address();
         let mut output_payments = PaymentsVec::new();
         for (payment, token_data) in payments.iter().zip(all_token_data.iter()) {
+            let token_balance = self.blockchain().get_esdt_balance(
+                &own_sc_address,
+                &payment.token_identifier,
+                payment.token_nonce,
+            );
+            if token_balance >= payment.amount {
```
CostinCarabas, line 89: If there are 2 payments with amount X and Y for the same token_id, then you save the 2 payments in `output_payments`, and
`X < token_balance`
`Y < token_balance`
but `X + Y > token_balance`.

multi-transfer-esdt/src/token_mapping.rs
dorin-iancu: We don't need the reverse storage. The bridge on the other side will handle this.

multi-transfer-esdt/src/transfer_tokens.rs
dorin-iancu: Fair, but how do I solve this easily?

PR #18, opened 2023-11-01
deposit function params and event update

psorinionut:


esdt-safe/src/to_sovereign/events.rs
```diff
@@ -38,7 +36,7 @@ pub trait EventsModule {
     fn deposit_event(
         &self,
         #[indexed] dest_address: &ManagedAddress,
-        #[indexed] tokens: &PaymentsVec<Self::Api>,
+        #[indexed] tokens: &MultiValueEncoded<MultiValue3<TokenIdentifier, u64, BigUint>>,
```
dorin-iancu, line 39: Don't use MultiValue types for events. It's bad practice, and doesn't do what you expect it to. Revert to PaymentsVec.

dorin-iancu (changes requested):

esdt-safe/src/to_sovereign/events.rs
sasurobert: @dorin-iancu can you fix this, as you work in this repo ?

esdt-safe/src/to_sovereign/create_tx.rs
```diff
@@ -14,29 +14,41 @@ pub trait CreateTxModule:
...
-    /// Create an Elrond -> Sovereign transaction.
+    /// Create an MultiversX -> Sovereign transaction.
     #[payable("*")]
     #[endpoint]
     fn deposit(
         &self,
         to: ManagedAddress,
-        opt_transfer_data: OptionalValue<TransferData<Self::Api>>,
+        opt_gas_limit: OptionalValue<GasLimit>,
+        opt_function: OptionalValue<ManagedBuffer>,
```
sasurobert, line 24: too many optionals - it is a bad practice. rewrite this. The thing is if we have gasLimit - we must have a function and we might have opt_args.

So make optional a structure which will have the GasLimit, Function and ArgsList.

sasurobert (changes requested):

sasurobert:
closed, as it was resolved in another PR.

PR #20, opened 2024-01-17
Fees Market

dorin-iancu:


fee-market/src/pairs.rs
```diff
@@ -0,0 +1,96 @@
...
+pub trait PairsModule {
+    #[only_owner]
+    #[endpoint(addPairs)]
+    fn add_pairs(
+        &self,
+        tokens_address_pairs: MultiValueEncoded<
+            MultiValue3<TokenIdentifier, TokenIdentifier, ManagedAddress>,
+        >,
+    ) {
+        for pairs in tokens_address_pairs {
```
sasurobert, line 27: I think here we need to integrate an oracle contracts, as the sovereign cross chain SCs will not stay in shard 1 where the DEX is.

sasurobert: we need a new interface and oracle contract which has all the prices. Maybe we could directly integrate here the oracle price contract, first update that, and then put it here.

sasurobert (approved):

PR #36, opened 2024-04-11
Alpha Sovereign SCs 

andreiblt1304:
This PR consists of the follow features:
1. alpha version of the Esdt-Safe contract
2. alpha version of the Fee-Market contract
3. alpha version of the Multisigverifier contract

Important notes:
- coverage and new proxies implementation for Safe-Price and SCs interactions will be added once Unified Syntax framework update is here


esdt-safe/src/to_sovereign/create_tx.rs
```diff
@@ -112,114 +115,205 @@ pub trait CreateTxModule:
...
-        opt_transfer_data: OptionalValue<TransferData<Self::Api>>,
-    ) {
+    #[endpoint(depositBack)]
+    fn deposit_back(&self, to: ManagedAddress) {
         require!(self.not_paused(), "Cannot create transaction while paused");
 
+        let caller = self.blockchain().get_caller();
+        require!(caller == ESDT_SYSTEM_SC_ADDRESS.into(), "Caller is invalid");
+
+        let payments = self.call_value().all_esdt_transfers();
```
alyn509, line 126: probably would be a good idea to keep the checks for payment not being empty and the number of tokens not exceeding maximum.

esdt-safe/src/to_sovereign/create_tx.rs
```diff
@@ -112,114 +115,205 @@ pub trait CreateTxModule:
...
+        opt_transfer_data: OptionalValue<
+            MultiValue3<GasLimit, ManagedBuffer, ManagedVec<ManagedBuffer>>,
+        >,
+    ) {
+        require!(self.not_paused(), "Cannot create transaction while paused");
+
+        let (fees_payment, payments) = self.check_and_extract_fee().into_tuple();
+
+        require!(!payments.is_empty(), "Nothing to transfer");
+        require!(payments.len() <= MAX_TRANSFERS_PER_TX, "Too many tokens");
```
alyn509, line 193: you can put these inside `check_and_extract_fee`

esdt-safe/src/to_sovereign/create_tx.rs
```diff
@@ -112,114 +115,205 @@ pub trait CreateTxModule:
...
+                    self.sovereign_esdt_token_info_mapper(&sov_token_id, &sov_token_nonce)
+                        .take();
+                }
+
+                event_payments.push(MultiValue3((
+                    sov_token_id,
+                    sov_token_nonce,
+                    current_token_data.clone(),
+                )));
             }
```
alyn509, line 279: Maybe you can extract some (or all) of this logic into some separate functions. Try to keep endpoint bodies smaller

alyn509:
general PR tips:

- when you work on multiple contracts, it would be better for both your work structure and for the reviewers to have a separate PR for each contract
- I started developing a hostility for commented tests when their contract was changes. I prefer a longer PR with tests included rather than to green this, merge and find out later in the tests that something is not right. But if the bullet above is met I think we are just ok
- endpoint bodies should be smaller and their core functionality split in multiple functions. This makes the logic easier to follow on review



P.S.: `multisigverifier/tests/multisigverifier_setup/mod.rs` is just dead code from test migration. It should be removed

esdt-safe/src/to_sovereign/create_tx.rs
andreiblt1304: done

esdt-safe/src/to_sovereign/create_tx.rs
andreiblt1304: done

esdt-safe/src/to_sovereign/create_tx.rs
andreiblt1304: I refactored the deposit endpoint

alyn509 (approved):

PR #56, opened 2024-06-20
ESDT Safe Cleanup

andreiblt1304:


mihaicalinluca (approved):

alyn509 (approved):

esdt-safe/src/to_sovereign/create_tx.rs
axenteoctavian, line 1: - the first 6 endpoints should be deleted
- put `deposit` endpoint as first function in this file

esdt-safe/src/from_sovereign/token_mapping.rs
```diff
@@ -12,54 +11,6 @@ pub struct EsdtTokenInfo<M: ManagedTypeApi> {
 pub trait TokenMappingModule:
     multiversx_sc_modules::default_issue_callbacks::DefaultIssueCallbacksModule
 {
-    #[payable("EGLD")]
-    #[endpoint(registerToken)]
-    fn register_token(
```
axenteoctavian, line 17: Please revert this change.
In `esdt-safe` contract we still use `register_token`. Only in `enshrine-esdt` we don't use it anymore.

esdt-safe/src/esdt_safe_proxy.rs
axenteoctavian, line 1: I'm not sure how this proxy file works but I have a question. 
None of the contracts have calls to esdt-safe, so do we need this proxy file?

axenteoctavian (changes requested):
Please resolve the comments

esdt-safe/src/to_sovereign/create_tx.rs
andreiblt1304: done

esdt-safe/src/to_sovereign/events.rs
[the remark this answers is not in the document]
andreiblt1304: deleted

esdt-safe/src/from_sovereign/token_mapping.rs
andreiblt1304: reverted

esdt-safe/src/esdt_safe_proxy.rs
andreiblt1304: We have a script to generate proxies for all the contracts, It is far easier to have all the proxies generated inside the src folder than to create them manually and then drag and drop to the correct path. Maybe in the future we will use that proxy.

axenteoctavian (approved):

PR #59, opened 2024-06-21
Enshrine ESDT Cleanup

andreiblt1304:


enshrine-esdt-safe/src/from_sovereign/token_mapping.rs
axenteoctavian, line 1: I don't think we need anything from this file anymore.

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
axenteoctavian, line 1: Please do the following:
- delete `mint_and_save_token` function
- on else condition at line 72 use the following code:

let token_data = operation_token.token_data.clone();
let mut arg_buffer = ManagedArgBuffer::new();
arg_buffer.push_arg(&operation_token.token_identifier);
arg_buffer.push_arg(token_data.amount);
arg_buffer.push_arg(token_data.name);
arg_buffer.push_arg(token_data.royalties);
arg_buffer.push_arg(token_data.hash);
arg_buffer.push_arg(token_data.attributes);
let uris = token_data.uris.clone();

if uris.is_empty() {
    // at least one URI is required, so we push an empty one
    arg_buffer.push_arg(codec::Empty);
} else {
    // The API function has the last argument as variadic,
    // so we top-encode each and send as separate argument
    for uri in &uris {
        arg_buffer.push_arg(uri);
    }
}
arg_buffer.push_arg(operation_token.token_nonce);
arg_buffer.push_arg(token_data.creator);

let output = self.send_raw().call_local_esdt_built_in_function(
    self.blockchain().get_gas_left(),
    &ManagedBuffer::from(ESDT_NFT_CREATE_FUNC_NAME),
    &arg_buffer,
);

if let Some(first_result_bytes) = output.try_get(0) {
    nonce = first_result_bytes.parse_as_u64().unwrap_or_default()
} else {
    nonce = 0
}

enshrine-esdt-safe/src/from_sovereign/token_mapping.rs
andreiblt1304: We do need code from here, including the mappers and the struct

axenteoctavian (approved):

PR #65, opened 2024-06-28
Wegld fee

andreiblt1304:
- Created a separate endpoint to register tokens
- Added registered tokens mapper
- Modified init endpoint with optional values for WEGLD and Sovereign Token Prefix Identifiers

### DIFF
--- a/common/utils/src/lib.rs
+++ b/common/utils/src/lib.rs
@@ -82,7 +82,9 @@ pub trait UtilsModule: bls_signature::BlsSignatureModule {
     }
 
     fn has_sov_prefix(&self, token_id: &TokenIdentifier, chain_prefix: ManagedBuffer) -> bool {
-        require!(self.has_prefix(token_id), "Token does not have prefix");
+        if !self.has_prefix(token_id) {
+            return false;
+        }
 
         let dash = b'-';
         let buffer = token_id.as_managed_buffer();
--- a/common/utils/src/lib.rs
+++ b/common/utils/src/lib.rs
@@ -80,4 +80,11 @@ pub trait UtilsModule: bls_signature::BlsSignatureModule {
 
         false
     }
+
+    #[inline]
+    fn has_sov_prefix(&self, token_id: &TokenIdentifier, sov_prefix: ManagedBuffer) -> bool {
+        require!(self.has_prefix(token_id), "Token does not have prefix");
+
+        *token_id == TokenIdentifier::from(sov_prefix)
+    }
 }
--- a/common/utils/src/lib.rs
+++ b/common/utils/src/lib.rs
@@ -7,6 +7,8 @@ multiversx_sc::imports!();
 pub type PaymentsVec<M> = ManagedVec<M, EsdtTokenPayment<M>>;
 
 static ERR_EMPTY_PAYMENTS: &[u8] = b"No payments";
+const DASH: u8 = b'-';
+const MAX_TOKEN_ID_LEN: usize = 32;
 
 #[multiversx_sc::module]
 pub trait UtilsModule: bls_signature::BlsSignatureModule {
@@ -67,12 +69,11 @@ pub trait UtilsModule: bls_signature::BlsSignatureModule {
     }
 
     fn has_prefix(&self, token_id: &TokenIdentifier) -> bool {
-        let dash = b'-';
         let buffer = token_id.as_managed_buffer();
-        let mut array_buffer = [0u8; 32];
+        let mut array_buffer = [0u8; MAX_TOKEN_ID_LEN];
         let slice = buffer.load_to_byte_array(&mut array_buffer);
 
-        let counter = slice.iter().filter(|&&c| c == dash).count();
+        let counter = slice.iter().filter(|&&c| c == DASH).count();
 
         if counter == 2 {
             return true;
@@ -86,12 +87,11 @@ pub trait UtilsModule: bls_signature::BlsSignatureModule {
             return false;
         }
 
-        let dash = b'-';
         let buffer = token_id.as_managed_buffer();
-        let mut array_buffer = [0u8; 32];
+        let mut array_buffer = [0u8; MAX_TOKEN_ID_LEN];
         let slice = buffer.load_to_byte_array(&mut array_buffer);
 
-        if let Some(index) = slice.iter().position(|&b| b == dash) {
+        if let Some(index) = slice.iter().position(|&b| b == DASH) {
             let prefix = ManagedBuffer::from(&slice[..index]);
 
             if prefix == chain_prefix {
--- a/common/utils/src/lib.rs
+++ b/common/utils/src/lib.rs
@@ -81,10 +81,22 @@ pub trait UtilsModule: bls_signature::BlsSignatureModule {
         false
     }
 
-    #[inline]
-    fn has_sov_prefix(&self, token_id: &TokenIdentifier, sov_prefix: ManagedBuffer) -> bool {
+    fn has_sov_prefix(&self, token_id: &TokenIdentifier, chain_prefix: ManagedBuffer) -> bool {
         require!(self.has_prefix(token_id), "Token does not have prefix");
 
-        *token_id == TokenIdentifier::from(sov_prefix)
+        let dash = b'-';
+        let buffer = token_id.as_managed_buffer();
+        let mut array_buffer = [0u8; 32];
+        let slice = buffer.load_to_byte_array(&mut array_buffer);
+
+        if let Some(index) = slice.iter().position(|&b| b == dash) {
+            let prefix = ManagedBuffer::from(&slice[..index]);
+
+            if prefix == chain_prefix {
+                return true;
+            }
+        }
+
+        false
     }
 }
--- a/common/utils/src/lib.rs
+++ b/common/utils/src/lib.rs
@@ -66,7 +66,7 @@ pub trait UtilsModule: bls_signature::BlsSignatureModule {
         list
     }
 
-    fn has_sov_token_prefix(&self, token_id: &TokenIdentifier) -> bool {
+    fn has_prefix(&self, token_id: &TokenIdentifier) -> bool {
         let dash = b'-';
         let buffer = token_id.as_managed_buffer();
         let mut array_buffer = [0u8; 32];
--- a/enshrine-esdt-safe/src/common/storage.rs
+++ b/enshrine-esdt-safe/src/common/storage.rs
@@ -7,4 +7,7 @@ pub trait CommonStorage {
 
     #[storage_mapper("wegldIdentifier")]
     fn wegld_identifier(&self) -> SingleValueMapper<TokenIdentifier>;
+
+    #[storage_mapper("sovereignTokenPrefix")]
+    fn sovereign_token_prefix(&self) -> SingleValueMapper<ManagedBuffer>;
 }
--- a/enshrine-esdt-safe/src/common/storage.rs
+++ b/enshrine-esdt-safe/src/common/storage.rs
@@ -8,6 +8,6 @@ pub trait CommonStorage {
     #[storage_mapper("wegldIdentifier")]
     fn wegld_identifier(&self) -> SingleValueMapper<TokenIdentifier>;
 
-    #[storage_mapper("sovereignTokenPrefix")]
+    #[storage_mapper("sovereignTokensPrefix")]
     fn sovereign_tokens_prefix(&self) -> SingleValueMapper<ManagedBuffer>;
 }
--- a/enshrine-esdt-safe/src/common/storage.rs
+++ b/enshrine-esdt-safe/src/common/storage.rs
@@ -9,5 +9,5 @@ pub trait CommonStorage {
     fn wegld_identifier(&self) -> SingleValueMapper<TokenIdentifier>;
 
     #[storage_mapper("sovereignTokenPrefix")]
-    fn sovereign_token_prefix(&self) -> SingleValueMapper<ManagedBuffer>;
+    fn sovereign_tokens_prefix(&self) -> SingleValueMapper<ManagedBuffer>;
 }
--- a/enshrine-esdt-safe/src/common/storage.rs
+++ b/enshrine-esdt-safe/src/common/storage.rs
@@ -6,5 +6,5 @@ pub trait CommonStorage {
     fn is_sovereign_chain(&self) -> SingleValueMapper<bool>;
 
     #[storage_mapper("wegldTicker")]
-    fn wegld_ticker(&self) -> SingleValueMapper<TokenIdentifier>;
+    fn wegld_identifier(&self) -> SingleValueMapper<TokenIdentifier>;
 }
--- a/enshrine-esdt-safe/src/common/storage.rs
+++ b/enshrine-esdt-safe/src/common/storage.rs
@@ -4,4 +4,7 @@ use multiversx_sc::imports::*;
 pub trait CommonStorage {
     #[storage_mapper("isSovereignChain")]
     fn is_sovereign_chain(&self) -> SingleValueMapper<bool>;
+
+    #[storage_mapper("wegldTicker")]
+    fn wegld_ticker(&self) -> SingleValueMapper<ManagedBuffer>;
 }
--- a/enshrine-esdt-safe/src/common/storage.rs
+++ b/enshrine-esdt-safe/src/common/storage.rs
@@ -6,5 +6,5 @@ pub trait CommonStorage {
     fn is_sovereign_chain(&self) -> SingleValueMapper<bool>;
 
     #[storage_mapper("wegldTicker")]
-    fn wegld_ticker(&self) -> SingleValueMapper<ManagedBuffer>;
+    fn wegld_ticker(&self) -> SingleValueMapper<TokenIdentifier>;
 }
--- a/enshrine-esdt-safe/src/common/storage.rs
+++ b/enshrine-esdt-safe/src/common/storage.rs
@@ -5,6 +5,6 @@ pub trait CommonStorage {
     #[storage_mapper("isSovereignChain")]
     fn is_sovereign_chain(&self) -> SingleValueMapper<bool>;
 
-    #[storage_mapper("wegldTicker")]
+    #[storage_mapper("wegldIdentifier")]
     fn wegld_identifier(&self) -> SingleValueMapper<TokenIdentifier>;
 }
--- a/enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
+++ b/enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
@@ -160,14 +160,16 @@ pub trait TransferTokensModule:
             self.register_token(token_identifier.clone_value());
         }
 
-        self.refund_wegld(sender, wegld_amount - wegld_fee_amount);
+        let remaining_wegld = wegld_amount - wegld_fee_amount;
+
+        if remaining_wegld > 0 {
+            self.refund_wegld(sender, remaining_wegld);
+        }
 
         return (registered_tokens, true);
     }
 
     fn refund_wegld(&self, sender: &ManagedAddress<Self::Api>, wegld_amount: BigUint<Self::Api>) {
-        require!(wegld_amount > 0, "No WEGLD fee sent");
-
         let wegld_identifier = self.wegld_identifier().get();
         let payment = EsdtTokenPayment::new(wegld_identifier, 0, wegld_amount);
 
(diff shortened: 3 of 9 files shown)


enshrine-esdt-safe/src/lib.rs
```diff
@@ -45,6 +45,13 @@ pub trait EnshrineEsdtSafe:
         self.header_verifier_address().set(&header_verifier_address);
     }
 
+    #[only_owner]
+    #[endpoint(setWegldTicker)]
+    fn set_wegld_ticker(&self, wegld_ticker: ManagedBuffer) {
```
axenteoctavian, line 50: I believe it's better to have the identifier in `init` as argument `Optional<ManagedBuffer>`.
It is only required for main chain contract, so you can do a check like this:
```
if is_sovereign_chain == false {
   //wegld_identifier should have value -> otherwise return error
}
```

enshrine-esdt-safe/src/common/storage.rs
```diff
@@ -4,4 +4,7 @@ use multiversx_sc::imports::*;
 pub trait CommonStorage {
     #[storage_mapper("isSovereignChain")]
     fn is_sovereign_chain(&self) -> SingleValueMapper<bool>;
+
+    #[storage_mapper("wegldTicker")]
```
axenteoctavian, line 8: Please use `wegldIdentifier` for storage name because it's the identifier, not the ticker

enshrine-esdt-safe/src/lib.rs
andreiblt1304: we can just use a `require!` to check for the `is_sovereign_chain` value to be true inside this function, this will also mean we don't have to use a `match` operator

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
```diff
@@ -36,15 +37,82 @@ pub trait TransferTokensModule:
...
+            call_payment.amount == DEFAULT_ISSUE_COST * tokens.len() as u64,
+            "WEGLD fee amount is not met"
+        );
+
+        for token in tokens {
+            self.paid_issued_tokens().insert(token);
+        }
+    }
+
+    fn check_tokens_for_wegld_fee(
```
axenteoctavian, line 77: This function is not really ok. 

Let's have some requirements first:
- input: `tokens`
- output: `wegld_amount` and `remainaing_tokens`

- if pos[0] is NOT WEGLD, verify all tokens don't have prefix or were already registered - meaning is transferring mainchain or registered tokens
- if pos[0] is WEGLD and len(tokens)=1 return `(0, tokens)` - meaning WEGLD was deposited
- if pos[0] is WEGLD and len(tokens)>1 and all tokens from pos[1..] do not have prefix or were already registered return `(0, tokens)` - meaning all tokens were registered or are from mainchain and WEGLD is on position 0 randomly
- if pos[0] is WEGLD and len(tokens)>1 and NOT all tokens from pos[1..] are registered return `(DEFAULT_ISSUE_COST * nr_of_unregisted, tokens[1..]) - meaning there are tokens where issue cost was not paid, sum wegld fee for all and return all tokens but WEGLD

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
```diff
@@ -36,15 +37,82 @@ pub trait TransferTokensModule:
             sc_panic!("Operation is not registered");
         }
 
-        let minted_operation_tokens = self.mint_tokens(&operation.tokens);
+        let (wegld_amount, checked_tokens) =
+            self.check_tokens_for_wegld_fee(operation.tokens.clone());
```
axenteoctavian, line 43: we can named this something like `verify_operation_tokens_for_issue_fee`

enshrine-esdt-safe/src/lib.rs
axenteoctavian: If you move this in `init` you can't use the require.

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
andreiblt1304: for the scenario where pos[0] is NOT WEGLD, should the function just return an error when there is a token that was not registered or it has a prefix ?

axenteoctavian: You need to return from function with some kind of error (maybe add a `bool` for output) and go through fail flow (send `execute_bridge_operation_event` and `deposit_event`)

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
```diff
@@ -36,13 +39,99 @@ pub trait TransferTokensModule:
...
+        &self,
+        sender: &ManagedAddress<Self::Api>,
+        tokens: ManagedVec<OperationEsdtPayment<Self::Api>>,
+    ) -> ManagedVec<OperationEsdtPayment<Self::Api>> {
+        let first_payment = tokens.get(0);
+        let is_first_payment_wegld = self.is_wegld(&first_payment.token_identifier);
+
+        if !is_first_payment_wegld {
+            for token in tokens.iter() {
+                if !self.was_token_registered(&token.token_identifier)
```
axenteoctavian, line 95: The condition should be `if token has prefix && is not registered then return ERROR` and at line 101 `return tokens`

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
```diff
@@ -36,13 +39,99 @@ pub trait TransferTokensModule:
...
+                    return ManagedVec::new();
+                }
+            }
+        }
+
+        if is_first_payment_wegld && tokens.len() == 1 {
+            return tokens;
+        }
+
+        let mut remaining_tokens = tokens.clone();
```
axenteoctavian, line 107: from here you should do it like this:
- for `token in tokens.iter().skip(1)` : `if has prefix && is not registered` add identifier in array
- if array empty -> `return tokens`
- else `wegld_amount_needed = DEFAULT_ISSUE_COST * len(array)`
-> if `tokens[0].amount >= wegld_amount_needed`: `register_token` from array, refund and return `tokens[1..]`
-> else return "ERROR not enough wegld"

enshrine-esdt-safe/src/lib.rs
```diff
@@ -24,9 +24,21 @@ pub trait EnshrineEsdtSafe:
     + common::storage::CommonStorage
 {
     #[init]
-    fn init(&self, is_sovereign_chain: bool) {
+    fn init(&self, is_sovereign_chain: bool, wegld_ticker: OptionalValue<ManagedBuffer>) {
```
axenteoctavian, line 27: `wegld_identifier`

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
```diff
@@ -36,13 +39,99 @@ pub trait TransferTokensModule:
...
+            };
+
+            self.distribute_payments(
+                hash_of_hashes.clone(),
+                operation_tuple,
+                minted_operation_tokens,
+            );
+        }
+
+        self.execute_bridge_operation_event(hash_of_hashes, operation_hash);
```
axenteoctavian, line 61: You have to use `emit_transfer_failed_events` here

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
andreiblt1304: I don't think it is necessary to create another array for this step, it's easier to register the identifier when it passes both conditions and keep a counter to calculate the fee afterwards

common/utils/src/lib.rs
```diff
@@ -80,4 +80,11 @@ pub trait UtilsModule: bls_signature::BlsSignatureModule {
 
         false
     }
+
+    #[inline]
+    fn has_sov_prefix(&self, token_id: &TokenIdentifier, sov_prefix: ManagedBuffer) -> bool {
```
axenteoctavian, line 85: let's use `sov_token: &TokenIdentifier` and `chain_prefix: ManagedBuffer`

common/utils/src/lib.rs
```diff
@@ -80,4 +80,11 @@ pub trait UtilsModule: bls_signature::BlsSignatureModule {
...
+        require!(self.has_prefix(token_id), "Token does not have prefix");
+
+        *token_id == TokenIdentifier::from(sov_prefix)
```
axenteoctavian, line 88: Here should be `prefix of sov_token == chain_prefix`

enshrine-esdt-safe/src/common/storage.rs
```diff
@@ -4,4 +4,10 @@ use multiversx_sc::imports::*;
 pub trait CommonStorage {
     #[storage_mapper("isSovereignChain")]
     fn is_sovereign_chain(&self) -> SingleValueMapper<bool>;
+
+    #[storage_mapper("wegldIdentifier")]
+    fn wegld_identifier(&self) -> SingleValueMapper<TokenIdentifier>;
+
+    #[storage_mapper("sovereignTokenPrefix")]
+    fn sovereign_token_prefix(&self) -> SingleValueMapper<ManagedBuffer>;
```
axenteoctavian, line 12: maybe use sovereign_token`s`_prefix

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
```diff
@@ -36,13 +39,139 @@ pub trait TransferTokensModule:
...
+        );
+
+        for token_id in tokens {
+            self.register_token(token_id);
+        }
+    }
+
+    #[endpoint(depositForNewRegistrations)]
+    #[payable("*")]
+    fn deposit_for_new_registrations(&self) {
```
axenteoctavian, line 95: This endpoint is not doing anything. No WEGLD check, no registration, no nothing...

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
```diff
@@ -36,13 +39,139 @@ pub trait TransferTokensModule:
...
+        let (wegld_amount, wegld_position) = self.get_wegld_payment_info(&tokens);
+
+        if wegld_position == -1 {
+            for token in tokens.iter() {
+                if !self.was_token_registered(&token.token_identifier) {
+                    return (ManagedVec::new(), false);
+                }
+            }
+
+            return (tokens, false);
```
axenteoctavian, line 132: This should return `(tokens, true)`

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
```diff
@@ -36,13 +39,139 @@ pub trait TransferTokensModule:
...
         };
 
-        self.distribute_payments(hash_of_hashes, operation_tuple, minted_operation_tokens);
+        if wegld_position >= 0 && tokens.len() == 1 {
```
axenteoctavian, line 135: - `wegld_position >= 0` is always true here, you can delete this condition.
- should return `(tokens, true)`

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
```diff
@@ -36,13 +39,139 @@ pub trait TransferTokensModule:
...
+            return (ManagedVec::new(), false);
+        }
+
+        for token_identifier in unregistered_tokens.iter() {
+            self.register_token(token_identifier.clone_value());
+        }
+
+        self.refund_wegld(sender, wegld_amount - wegld_fee_amount);
+
+        return (registered_tokens, false);
```
axenteoctavian, line 165: return `(registered_tokens, true)`

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
```diff
@@ -36,13 +39,139 @@ pub trait TransferTokensModule:
...
+    }
+
+    fn refund_wegld(&self, sender: &ManagedAddress<Self::Api>, wegld_amount: BigUint<Self::Api>) {
+        require!(wegld_amount > 0, "No WEGLD fee sent");
```
axenteoctavian, line 169: This `require!` isn't stopping the transaction execution?

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
```diff
@@ -52,7 +181,10 @@ pub trait TransferTokensModule:
         let mut output_payments = ManagedVec::new();
 
         for operation_token in operation_tokens.iter() {
-            if !self.has_sov_token_prefix(&operation_token.token_identifier) {
+            if !self.has_sov_prefix(
```
axenteoctavian, line 184: You have to use `self.has_prefix` here because you can mint any token with prefix

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
```diff
@@ -208,6 +340,18 @@ pub trait TransferTokensModule:
             .sync_call();
     }
 
+    fn burn_sovereign_tokens(&self, operation: &Operation<Self::Api>) {
+        for token in operation.tokens.iter() {
+            if self.has_sov_prefix(&token.token_identifier, self.get_sovereign_prefix()) {
```
axenteoctavian, line 345: To be consistent with mint function, you can use
`if !self.has_prefix { continue }` otherwise burn

enshrine-esdt-safe/src/lib.rs
```diff
@@ -24,9 +24,35 @@ pub trait EnshrineEsdtSafe:
...
+        &self,
+        is_sovereign_chain: bool,
+        wegld_identifier: Option<ManagedBuffer>,
+        sov_token_prefix: Option<ManagedBuffer>,
+    ) {
         self.is_sovereign_chain().set(is_sovereign_chain);
         self.set_paused(true);
+
+        if !is_sovereign_chain {
+            match (wegld_identifier, sov_token_prefix) {
```
axenteoctavian, line 37: This looks very hard to maintain if you have a lot of arguments. 
- it would be better to use `OptionalValue` instead of `Option`
- just take the arguments one by one and `sc_panic!` if error
- don't forget to throw error if wegld identifier is invalid

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
```diff
@@ -36,13 +39,139 @@ pub trait TransferTokensModule:
...
+        (BigUint::from(0u32), -1)
+    }
+
+    fn verify_operation_tokens_for_issue_fee(
+        &self,
+        sender: &ManagedAddress<Self::Api>,
+        tokens: ManagedVec<OperationEsdtPayment<Self::Api>>,
+    ) -> (ManagedVec<OperationEsdtPayment<Self::Api>>, bool) {
+        require!(!tokens.is_empty(), "Tokens array should not be empty");
+        let (wegld_amount, wegld_position) = self.get_wegld_payment_info(&tokens);
```
axenteoctavian, line 123: If you can have WEGLD on any position, How do you handle a case where I send 0.05 WEGLD for register but I also want to send some WEGLD on the receiver?
```
1 sov-ESDT-12345
1 WEGLD
0.05 WEGLD
```

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
andreiblt1304: My thoughts are that I either remove it altogether or I create another storage mapper that holds the current amount of WEGLD in the pool

enshrine-esdt-safe/src/lib.rs
andreiblt1304: you can't use 2 OptionalValues as parameters

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
andreiblt1304: I didn't think of this scenario, maybe we can add all the wegld amounts to 1 variable and then subtract the needed amount and refund the rest?

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
```diff
@@ -36,13 +39,66 @@ pub trait TransferTokensModule:
...
 
-        self.distribute_payments(hash_of_hashes, operation_tuple, minted_operation_tokens);
+        self.distribute_payments(
+            hash_of_hashes.clone(),
+            operation_tuple,
+            minted_operation_tokens,
+        );
+    }
+
+    #[endpoint(registerTokens)]
```
axenteoctavian, line 68: Let's use the documentation naming `registerNewTokenID`

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
```diff
@@ -36,13 +39,66 @@ pub trait TransferTokensModule:
...
+        for token_id in tokens {
+            self.register_token(token_id);
+        }
+    }
+
+    fn verify_operation_tokens_issue_paid(
+        &self,
+        tokens: ManagedVec<OperationEsdtPayment<Self::Api>>,
+    ) -> bool {
+        require!(!tokens.is_empty(), "Tokens array should not be empty");
```
axenteoctavian, line 93: You said MultiESDTNFTTransfer cannot have 0 tokens as payment so this require is not needed. Also this require should never happen because will fail the transactions and not send the fail events

axenteoctavian (approved):

axenteoctavian (approved):

common/utils/src/lib.rs
```diff
@@ -80,4 +80,25 @@ pub trait UtilsModule: bls_signature::BlsSignatureModule {
 
         false
     }
+
+    fn has_sov_prefix(&self, token_id: &TokenIdentifier, chain_prefix: ManagedBuffer) -> bool {
+        if !self.has_prefix(token_id) {
+            return false;
+        }
+
+        let dash = b'-';
```
dorin-iancu, line 89: Have this as a const instead.

common/utils/src/lib.rs
```diff
@@ -80,4 +80,25 @@ pub trait UtilsModule: bls_signature::BlsSignatureModule {
...
+        let buffer = token_id.as_managed_buffer();
+        let mut array_buffer = [0u8; 32];
```
dorin-iancu, line 91: "32" can also be a const, like "MAX_TOKEN_ID_LEN"

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
```diff
@@ -36,13 +39,70 @@ pub trait TransferTokensModule:
...
+            "WEGLD is the only token accepted as register fee"
+        );
+
+        require!(
+            call_payment.amount == DEFAULT_ISSUE_COST * tokens.len() as u64,
+            "WEGLD fee amount is not met"
+        );
+
+        for token_id in tokens {
+            self.register_token(token_id);
```
dorin-iancu, line 87: If this makes token issue, it won't work. You can only make one legacy async call per call, which is what the token issue framework uses.

enshrine-esdt-safe/src/lib.rs
```diff
@@ -24,9 +24,34 @@ pub trait EnshrineEsdtSafe:
...
+                }
+
+                None => sc_panic!("WEGLG identifier must be set in Mainchain"),
+            }
+
+            match sov_token_prefix {
+                Some(prefix) => self.sovereign_tokens_prefix().set(prefix),
+                None => sc_panic!("Sovereign Token Prefix must be set in Mainchain"),
+            }
+        }
```
dorin-iancu, line 54: Can revert condition here to do some de-nesting, i.e.
```
if is_sovereign_chain {
    return;
}

// code that was inside if
```

Also, since you need "Some" for both Option, why not take "TokenIdentifier" and "ManagedBuffer" directly in the init call?

common/utils/src/lib.rs
dorin-iancu: Should also do the same in the above function btw.

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
andreiblt1304: `register_token` just saves the `TokenIdentifier` to the `paid_issued_tokens` mapper

enshrine-esdt-safe/src/lib.rs
andreiblt1304: i agree with the de-nesting but I don't quite understand what you mean by taking those directly in the init call

alyn509 (approved):

PR #73, opened 2024-07-25
Distribute Payments

andreiblt1304:


token-handler/src/mint_tokens.rs
```diff
@@ -100,6 +82,41 @@ pub trait MintTokensModule:
...
     }
 
+    fn prepare_nft_create_args(
+        &self,
+        token_identifier: &TokenIdentifier<Self::Api>,
+        token_nonce: &u64,
+        token_data: &StolenFromFrameworkEsdtTokenData<Self::Api>,
+    ) -> ManagedArgBuffer<Self::Api> {
+        let mut arg_buffer = ManagedArgBuffer::new();
+        let cloned_token_data = token_data.clone();
```
andreiblt1304, line 92: @mihaicalinluca do you think there is a more efficient way of using this reference without cloning it ? `push_arg` consumes the parameter

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
```diff
@@ -46,23 +42,35 @@ pub trait TransferTokensModule:
...
+        self.tx()
+            .to(token_handler_address)
+            .typed(token_handler_proxy::TokenHandlerProxy)
+            .mint_tokens(hash_of_hashes, OperationTuple { op_hash, operation })
+            .sync_call();
     }
 
     #[endpoint]
-    fn call_token_handler_mint_endpoint(&self, operation: Operation<Self::Api>) {
+    fn call_token_handler_mint_endpoint(
```
axenteoctavian, line 63: What is this endpoint? Why do you need this?
What is `save_minted_tokens` callback at line 74?

token-handler/Cargo.toml
```diff
@@ -17,6 +17,12 @@ path = "../common/transaction"
 [dependencies.bls-signature]
 path = "../common/bls-signature/"
 
+[dependencies.header-verifier]
+path = "../header-verifier/"
+
+[dependencies.tx-batch-module]
+path = "../common/tx-batch-module"
```
axenteoctavian, line 24: I don't think we use this module anymore `tx-batch-module`

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
andreiblt1304: i used it in some random tests before implementing the distribute payments inside the token-handler contract, should I remove it here or in another PR ?

token-handler/src/burn_tokens.rs
axenteoctavian, line 1: I don't think we will use this burn module. With `transfer and execute by user` if the transaction fails, tokens will be sent to the original_sender wallet in main chain automatically, so we will not burn tokens. You can just keep it here for the moment.

token-handler/Cargo.toml
andreiblt1304: we do, we use it in `emit_transfer_failed_events` function 
`let tx_nonce = self.get_and_save_next_tx_id();`

token-handler/src/common/events.rs
```diff
@@ -0,0 +1,21 @@
...
+pub trait EventsModule {
+    #[event("executedBridgeOp")]
+    fn execute_bridge_operation_event(
+        &self,
+        #[indexed] hash_of_hashes: ManagedBuffer,
+        #[indexed] hash_of_bridge_op: ManagedBuffer,
+    );
+
+    #[event("deposit")]
+    fn deposit_event(
```
axenteoctavian, line 15: This event should be in enshrine, token-handler will not emit this event (after transfer and execute by user)

token-handler/src/common/storage.rs
```diff
@@ -4,4 +4,7 @@ use multiversx_sc::imports::SingleValueMapper;
 pub trait CommonStorage {
     #[storage_mapper]
     fn sov_prefix(&self) -> SingleValueMapper<ManagedBuffer>;
+
+    #[storage_mapper]
+    fn header_verifier_address(&self) -> SingleValueMapper<ManagedAddress>;
```
axenteoctavian, line 9: In header-verifier we have a condition that only enshrine is able to call remove_executed_hash, so this storage needs to be in enshrined.

We can change the implementation that only token-handler can remove_executed_hash, but we need to analyse if this is safe. Anyways this would be another PR

andreiblt1304: @sasurobert do you think that `token-handler` should remove the executed hashes from the `header-verifier` or keep that role to `enshrine-esdt` ?

token-handler/src/lib.rs
```diff
@@ -5,19 +5,27 @@ use multiversx_sc::imports::*;
...
+    mint_tokens::MintTokensModule
+    + burn_tokens::BurnTokensModule
     + utils::UtilsModule
     + common::storage::CommonStorage
+    + tx_batch_module::TxBatchModule
+    + common::events::EventsModule
 {
     #[init]
-    fn init(&self, chain_prefix: ManagedBuffer) {
+    fn init(&self, header_verifier_address: ManagedAddress, chain_prefix: ManagedBuffer) {
```
axenteoctavian, line 21: The 2 arguments are wrong.
I guess you need header_verifier_address this to call remove_executed_hash but this is wrong because 1) there are multiple header verifier contracts (one for each enshrine) and 2) remove_executed_hash has a condition to be called only by one specific enshrine.
Same problem for chain_prefix, there are multiple enshrine contracts that call one token-handler contract (in each shard)

token-handler/src/mint_tokens.rs
```diff
@@ -1,58 +1,63 @@
...
+pub trait MintTokensModule:
+    utils::UtilsModule
+    + common::storage::CommonStorage
+    + common::events::EventsModule
+    + burn_tokens::BurnTokensModule
+    + tx_batch_module::TxBatchModule
+{
     #[endpoint(mintTokens)]
-    fn mint_tokens(&self, operation_tokens: MultiValueEncoded<OperationEsdtPayment<Self::Api>>) {
+    fn mint_tokens(
```
axenteoctavian, line 27: This should not be called mint_tokens, should be something like `transfer_tokens` maybe.
It does: 
1) mint tokens
2) distribute tokens

token-handler/src/lib.rs
andreiblt1304: i think a good idea is to create a storage mapper dictionary so we have the associated addresses:
HeaderVerifier A <-> Enshrine A

token-handler/src/mint_tokens.rs
```diff
@@ -1,58 +1,63 @@
...
+                    .to(ToSelf)
+                    .typed(system_proxy::UserBuiltinProxy)
+                    .esdt_local_mint(
+                        &operation_token.token_identifier,
+                        operation_token.token_nonce,
+                        &operation_token.token_data.amount,
+                    )
+                    .sync_call();
+            } else {
+                let arg_buffer = self.get_nft_create_args(
```
axenteoctavian, line 56: Better to just make the function to `create_args` and `mint` and return the nonce

token-handler/src/mint_tokens.rs
andreiblt1304: I will rename the module as well to `TransferTokensModule`

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
```diff
@@ -46,23 +42,35 @@ pub trait TransferTokensModule:
...
                     operation: operation.clone(),
                 },
             );
 
             return;
         }
+
+        let token_handler_address = self.token_handler_address().get();
+
+        self.tx()
```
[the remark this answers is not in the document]
andreiblt1304: if the token handler will mint the tokens shouldn't I send the tokens WITH the prefix in the transaction ?

token-handler/src/common/storage.rs
axenteoctavian: I would say to not complicate `token-handler` contract with whitelist header verifier contracts and stuff. 
`enshrine-esdt` is verifying hash is present in `header_verifier` so enshrine will also remove it in the end.

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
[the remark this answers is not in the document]
axenteoctavian: You need to transfer tokens without prefix, and send a list of tokens with prefix as argument to be minted by the contract

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
andreiblt1304: deleted

token-handler/src/common/storage.rs
sasurobert: i agree with octavian.

token-handler/src/lib.rs
sasurobert: I also think a storage mapper dictionary is needed.

Separation of contracts is a must, so every enshrined sovereign must have its own set of contracts.

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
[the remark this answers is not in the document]
sasurobert: @axenteoctavian why do you say that transfers should happen without prefix ?

On the mainnet, it is ok to have prefixed tokens.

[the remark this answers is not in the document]
axenteoctavian: Sorry. 
I mean tokens "saved" (without prefix or with prefix != the one from enshrine) will pe transferred from enshrine to token-handler, and the rest of them will be minted by token-handler, then all will be distributed

token-handler/src/mint_tokens.rs
mihaicalinluca: from my testing you can just use references there as such:

```
arg_buffer.push_arg(&token_identifier);
        arg_buffer.push_arg(&token_data.amount);
        arg_buffer.push_arg(&token_data.name);
        arg_buffer.push_arg(&token_data.royalties);
        arg_buffer.push_arg(&token_data.hash);
        arg_buffer.push_arg(&token_data.attributes);
```

token-handler/src/lib.rs
axenteoctavian: - Yes, each Enshrine has a HeaderVerifier storage inside. TokenHandler only needs a list of Enshrine contracts, which is implemented [here](https://github.com/multiversx/mx-sovereign-sc/pull/76).
- TokenHandler is not connected to one Enshrine (to one prefix), is connected to all Enshrine from its shard. TokenHandler can mint any token with any prefix

token-handler/Cargo.toml
axenteoctavian: The module is deprecated, later you can do another PR to delete it and move the nonce thing somewhere else

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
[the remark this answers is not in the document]
andreiblt1304: so the ones that are already in the storage mapper will be sent with a transfer and the others will be send with the endpoint call, correct ?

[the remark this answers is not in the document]
axenteoctavian: There is no storage mapper. You look in the `Operation` tokens.

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
```diff
@@ -107,204 +87,45 @@ pub trait TransferTokensModule:
...
-        tokens: ManagedVec<OperationEsdtPayment<Self::Api>>,
-    ) -> bool {
+        tokens: &ManagedVec<OperationEsdtPayment<Self::Api>>,
+    ) -> (
+        ManagedVec<Self::Api, EsdtTokenPayment<Self::Api>>,
+        ManagedVec<Self::Api, OperationEsdtPayment<Self::Api>>,
+        bool,
+    ) {
         let sov_prefix = self.get_sovereign_prefix();
+        let mut sov_tokens: ManagedVec<Self::Api, EsdtTokenPayment<Self::Api>> = ManagedVec::new();
```
axenteoctavian, line 99: `sov_tokens` should be `OperationEsdtPayment` because these will be minted.
`non_sov_tokens` should be `EsdtTokenPayment` because these will be transferred from enshrine to token-handler as payment

enshrine-esdt-safe/src/from_sovereign/transfer_tokens.rs
```diff
@@ -32,58 +29,41 @@ pub trait TransferTokensModule:
...
-            .mint_tokens(MultiValueEncoded::from(operation.tokens.clone()))
-            .callback(<Self as TransferTokensModule>::callbacks(self).save_minted_tokens())
-            .async_call_and_exit();
-    }
+            .transfer_tokens(
+                operation.data.opt_transfer_data,
+                operation.to,
+                multi_value_tokens,
+            )
+            .multi_esdt(sov_tokens)
```
axenteoctavian, line 62: `non_sov_tokens` should be payment and `sov_tokens` should be in arguments for `transfer_tokens`

token-handler/src/common/events.rs
axenteoctavian, line 1: `token-handler` contract is not transmitting any events, this file can be deleted

token-handler/src/lib.rs
```diff
@@ -5,16 +5,17 @@ use multiversx_sc::imports::*;
...
-    + burn_tokens::BurnTokens
+    mint_tokens::TransferTokensModule
+    + burn_tokens::BurnTokensModule
     + utils::UtilsModule
     + common::storage::CommonStorage
+    + tx_batch_module::TxBatchModule
+    + common::events::EventsModule
 {
     #[init]
     fn init(&self, chain_prefix: ManagedBuffer) {
```
axenteoctavian, line 21: The `chain-prefix` argument should not exist for this contract.

token-handler/src/mint_tokens.rs
```diff
@@ -1,79 +1,205 @@
...
+    + common::storage::CommonStorage
+    + common::events::EventsModule
+    + burn_tokens::BurnTokensModule
+    + tx_batch_module::TxBatchModule
+{
+    #[payable("*")]
+    #[endpoint(transferTokens)]
+    fn transfer_tokens(
+        &self,
+        opt_transfer_data: Option<TransferData<Self::Api>>,
```
axenteoctavian, line 28: `Option` argument should be the last one?

token-handler/src/mint_tokens.rs
```diff
@@ -1,79 +1,205 @@
...
+            // at least one URI is required, so we push an empty one
+            arg_buffer.push_arg(codec::Empty);
+        } else {
+            // The API function has the last argument as variadic,
+            // so we top-encode each and send as separate argument
+            for uri in &uris {
+                arg_buffer.push_arg(uri);
+            }
+        }
+
```
axenteoctavian, line 118: We need one more argument here (before nonce): arg_buffer.push_arg(cloned_token_data.token_type);

token-handler/src/mint_tokens.rs
```diff
@@ -1,79 +1,205 @@
...
-            // self.minted_tokens().push(&operation_token);
-
             output_payments.push(OperationEsdtPayment {
                 token_identifier: operation_token.token_identifier,
                 token_nonce: nonce,
                 token_data: operation_token.token_data,
             });
         }
+
+        self.distribute_payments(&tokens_vec, &opt_transfer_data, &to);
```
axenteoctavian, line 73: You distribute `output_payments` (the minted tokens) plus `self.call_value().all_esdt_transfers();`

token-handler/src/mint_tokens.rs
```diff
@@ -1,79 +1,205 @@
...
+            }
+        }
+
+        arg_buffer.push_arg(token_nonce);
+        arg_buffer.push_arg(cloned_token_data.creator);
+
+        arg_buffer
+    }
+
+    fn distribute_payments(
```
axenteoctavian, line 125: Reorder arguments: (receiver, tokens, opt_transfer_data)

token-handler/src/mint_tokens.rs
```diff
@@ -1,79 +1,205 @@
...
+        for token in mapped_tokens {
+            args.push_arg(token.token_identifier);
+            args.push_arg(token.token_nonce);
+            args.push_arg(token.amount);
+        }
+
+        args
+    }
+
+    fn emit_transfer_failed_events(
```
axenteoctavian, line 181: This shouldn't be here and is not used, you can delete it.

token-handler/src/mint_tokens.rs
```diff
@@ -1,79 +1,205 @@
...
+pub trait TransferTokensModule:
+    utils::UtilsModule
+    + common::storage::CommonStorage
+    + common::events::EventsModule
+    + burn_tokens::BurnTokensModule
+    + tx_batch_module::TxBatchModule
+{
+    #[payable("*")]
+    #[endpoint(transferTokens)]
+    fn transfer_tokens(
```
axenteoctavian, line 26: We will also need the `original_sender` (operation.data.op_sender) in argument. This will be later used for transfer and execute by user functionality

token-handler/src/mint_tokens.rs
andreiblt1304: `MultiValueEncoded` should be the last one

token-handler/src/mint_tokens.rs
```diff
@@ -1,79 +1,187 @@
...
+        &self,
+        opt_transfer_data: Option<TransferData<Self::Api>>,
+        to: ManagedAddress,
+        // original_sender: ManagedAddress,
+        tokens: MultiValueEncoded<OperationEsdtPayment<Self::Api>>,
+    ) {
+        let mut output_payments = self.mint_tokens(&tokens.to_vec());
```
axenteoctavian, line 30: Let's make `mint_tokens` function to return array of `ManagedVec<EsdtTokenPayment>`, then extend with `call_value_esdt_transfer`. And use this in `distribute_payments`.

token-handler/src/mint_tokens.rs
```diff
@@ -1,79 +1,187 @@
...
+                    &operation_token.token_nonce,
+                    &operation_token.token_data,
                 );
 
-                if let Some(first_result_bytes) = output.try_get(0) {
-                    nonce = first_result_bytes.parse_as_u64().unwrap_or_default()
-                } else {
-                    nonce = 0
-                }
+                nonce = self.call_nft_create_built_in_function(&arg_buffer);
```
axenteoctavian, line 69: I believe we don't need `nonce` variable because it will always be `operation_token.token_nonce`. We can use here the mint function directly:

```
self.send_raw().call_local_esdt_built_in_function(
    self.blockchain().get_gas_left(),
    &ManagedBuffer::from(ESDT_NFT_CREATE_FUNC_NAME),
    arg_buffer,
);
```

token-handler/src/mint_tokens.rs
axenteoctavian, line 1: Let's call this file `transfer_tokens.rs`.

token-handler/src/transfer_tokens.rs
```diff
@@ -0,0 +1,168 @@
...
+                );
+            }
+
+            let esdt_token_payment = EsdtTokenPayment::new(
+                operation_token.token_identifier,
+                operation_token.token_nonce,
+                operation_token.token_data.amount,
+            );
+
+            output_payments.push(esdt_token_payment);
```
axenteoctavian, line 74: I believe you can use `output_payments.push(operation_token.into());`

token-handler/src/burn_tokens.rs
axenteoctavian, line 1: Let's delete this `burn_tokens.rs` module because it's not used. Also, delete the `token-handler/src/common` because will not be needed anymore

token-handler/src/storage.rs
axenteoctavian, line 1: Please delete this file, it's not used

sasurobert (approved):

axenteoctavian (approved):

PR #76, opened 2024-07-26
Enshrine ESDT address whitelist

andreiblt1304:


token-handler/src/lib.rs
```diff
@@ -23,4 +23,25 @@ pub trait TokenHandler:
 
     #[upgrade]
     fn upgrade(&self) {}
+
+    #[only_owner]
+    #[endpoint(setEnshrineEsdtWhitelist)]
+    fn set_enshrine_esdt_whitelist(
```
axenteoctavian, line 29: This should be `whitelist_enshrine_esdt` and pass only one contract address.
Also add a TODO to make this endpoint callable only by factorySC or configSC, I'm not sure which one.

token-handler/src/mint_tokens.rs
```diff
@@ -1,79 +1,205 @@
...
+    + tx_batch_module::TxBatchModule
+{
+    #[payable("*")]
+    #[endpoint(transferTokens)]
+    fn transfer_tokens(
+        &self,
+        opt_transfer_data: Option<TransferData<Self::Api>>,
+        to: ManagedAddress,
+        tokens: MultiValueEncoded<OperationEsdtPayment<Self::Api>>,
+    ) {
```
axenteoctavian, line 31: You have to `require!` that caller is whitelisted first thing in this function

token-handler/src/lib.rs
```diff
@@ -3,14 +3,28 @@
...
+pub trait TokenHandler:
+    transfer_tokens::TransferTokensModule + common_storage::CommonStorage
+{
     #[init]
     fn init(&self) {}
 
     #[upgrade]
     fn upgrade(&self) {}
+
+    #[only_owner]
```
axenteoctavian, line 20: This is not ok because the owner of the contract will probably be a multisig or something.

I think the caller for this endpoint should be FactorySC from same shard (this SC will deploy sovereign cross chain contracts so he can whitelist the esdt-safe address). Maybe whitelist the Factory SC address in `init`. What do you think @sasurobert ?

alyn509: you can try admin-only and add the address of the SC you want to call this as admin

sasurobert: We can set the address on the init as well.

I think it is good now as well. The owner of all contracts will be the factorySC.

alyn509 (approved):

axenteoctavian (approved):

PR #82, opened 2024-08-04
`addContractsToMap` endpoint

andreiblt1304:
Endpoint: addContractsToMap@list<contractName, address>
The initial setup of the contract will be made by a user account and finally it will give the ownership to the controller. On initial setup we will add the map of address to base contract and base contracts to address. This is needed in order to use deployContractFromSource/upgradeContractFromSource. The list of needed contracts are: chainFactorySC, controllerSC, sovereignHeaderVerifierSC, sovereignCrossChainOperationSC, chainConfigSC, slashingSC.


chain-factory/src/factory.rs
```diff
@@ -42,9 +61,38 @@ pub trait FactoryModule {
...
+            require!(
+                self.is_required_contract_name(&contract.name),
+                "Given contract name is not a required SC"
+            );
+
+            self.contracts_map(contract.name).set(contract.address);
+        }
+    }
+
+    fn is_required_contract_name(&self, contract_name: &ManagedBuffer) -> bool {
```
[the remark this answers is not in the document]
andreiblt1304: `#[inline]` is a rust annotation that it's used for efficiency, basically instead of calling the function it actually replaces the call with the body of the function
https://matklad.github.io/2021/07/09/inline-in-rust.html
At the moment I don't think it's necessary to inline it since it maps over the `SC_ARRAY`

alyn509 (approved):

chain-factory/src/factory.rs
```diff
@@ -42,9 +61,39 @@ pub trait FactoryModule {
...
+        for contract in contracts_list {
+            require!(
+                self.blockchain().is_smart_contract(&contract.address),
+                "Given contract address is not a correct SC address"
+            );
+
+            require!(
+                self.is_required_contract_name(&contract.name),
+                "Given contract name is not a required SC"
+            );
```
dorin-iancu, line 79: Instead of returning bool and checking, simply name the function `require_known_contract_name` and do the require there.

chain-factory/src/factory.rs
```diff
@@ -42,9 +61,39 @@ pub trait FactoryModule {
...
+
+            self.contracts_map(contract.name).set(contract.address);
```
dorin-iancu, line 81: Maybe check `is_empty` to not overwrite?

```
let mapper = self.contracts_map(contract.name);
require!(mapper.is_empty(), "Contract already set");

mapper.set(contract.address);
```

chain-factory/src/factory.rs
```diff
@@ -42,9 +61,39 @@ pub trait FactoryModule {
         let _ = self.all_deployed_contracts().swap_remove(&sc_address);
     }
 
+    #[only_owner]
+    #[endpoint(addContractsToMap)]
+    fn add_contracts_to_map(
+        &self,
+        contracts_list: MultiValueEncoded<Self::Api, ContractMap<Self::Api>>,
```
dorin-iancu, line 68:
```suggestion
        contracts_list: MultiValueEncoded<ContractMap<Self::Api>>,
```

chain-factory/src/factory.rs
```diff
@@ -1,7 +1,25 @@
...
 
 multiversx_sc::imports!();
 
+#[derive(TopDecode, TypeAbi, ManagedVecItem, Clone)]
+pub struct ContractMap<M: ManagedTypeApi> {
+    pub name: ManagedBuffer<M>,
+    pub address: ManagedAddress<M>,
+}
+
+const SC_ARRAY: [&str; 6] = [
```
dorin-iancu, line 14: Might want to expose these to a view so they don't have to scavenge through code to find them.

Also, for efficiency, use `&[u8]` instead of `&str", and add a `b` in front of the strings.

andreiblt1304: but how can I check if the contracts name are the correct ones if I put them inside a view ?

PR #91, opened 2024-08-07
code changes

dorin-iancu:


alyn509 (approved):

esdt-safe/src/from_sovereign/token_mapping.rs
```diff
@@ -28,28 +46,31 @@ pub trait TokenMappingModule:
...
                 token_type,
                 issue_cost,
                 token_display_name,
                 token_ticker,
                 num_decimals,
-            ),
+            }),
         }
 
+        // !!! Unreachable code !!!
```
axenteoctavian, line 73: Really? This worked before :thinking:

andreiblt1304: I was thinking the same thing

andreiblt1304: does it really matter if we will probably won't use `esdt-safe`?

axenteoctavian: Currently this contact is used for public deployment. Hopefully soon we'll move to enshrine-esdt.

axenteoctavian (approved):

esdt-safe/src/from_sovereign/token_mapping.rs
dorin-iancu: Yep. Issue-ing a token kills current contract execution.

PR #108, opened 2024-08-20
Add contracts to map

andreiblt1304:
Endpoint: addContractsToMap@list<contractName, address>
The initial setup of the contract will be made by a user account and finally it will give the ownership to the controller. On initial setup we will add the map of address to base contract and base contracts to address. This is needed in order to use deployContractFromSource/upgradeContractFromSource. The list of needed contracts are: chainFactorySC, controllerSC, sovereignHeaderVerifierSC, sovereignCrossChainOperationSC, chainConfigSC, slashingSC.

### DIFF
--- a/chain-factory/src/factory.rs
+++ b/chain-factory/src/factory.rs
@@ -3,20 +3,24 @@ use chain_config::StakeMultiArg;
 multiversx_sc::derive_imports!();
 multiversx_sc::imports!();
 
-#[derive(TopEncode, TopDecode, TypeAbi, Clone, ManagedVecItem)]
+#[derive(TypeAbi, TopEncode, TopDecode, NestedEncode, NestedDecode, ManagedVecItem)]
 struct ContractMapArgs<M: ManagedTypeApi> {
-    name: ManagedBuffer<M>,
+    name: ScArray,
     address: ManagedAddress<M>,
 }
 
-const SC_ARRAY: &[&[u8]] = &[
-    b"chainFactory",
-    b"controller",
-    b"sovereignHeaderVerifier",
-    b"sovereignCrossChainOperation",
-    b"chainConfig",
-    b"slashing",
-];
+#[derive(
+    TypeAbi, TopEncode, TopDecode, NestedEncode, NestedDecode, Clone, ManagedVecItem, PartialEq,
+)]
+pub enum ScArray {
+    None,
+    ChainFactory,
+    Controller,
+    SovereignHeaderVerifier,
+    SovereignCrossChainOperation,
+    ChainConfig,
+    Slashing,
+}
 
 #[multiversx_sc::module]
 pub trait FactoryModule {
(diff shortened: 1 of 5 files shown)


axenteoctavian (approved):

chain-factory/src/factory.rs
```diff
@@ -60,16 +64,13 @@ pub trait FactoryModule {
     ) {
         require!(!contracts_map.is_empty(), "Given contracts map is empty");
 
-        let mapped_array: ManagedVec<ManagedBuffer> = SC_ARRAY
-            .into_iter()
-            .map(|sc| ManagedBuffer::from(*sc))
-            .collect();
+        for contract in contracts_map {
+            require!(
+                contract.name != ScArray::None,
```
axenteoctavian, line 69: 1. Why do you need `ScArray::None`?
2. What happens if endpoint receives index 100?
3. If you use enum maybe use `contract.id` instead of `contract.name`

andreiblt1304: The None is used so the index can start from 1, if it receives index 100 the transaction will fail.

axenteoctavian: What is the problem if index 0 is a contract?

andreiblt1304: The problem is that using index 0 in a storage mapper will not store anything inside the contract, and since the storage mapper `contracts_map` uses an index for the ID, it won't do anything for ID 0

chain-factory/src/factory.rs
```diff
@@ -1,7 +1,27 @@
...
+struct ContractMapArgs<M: ManagedTypeApi> {
+    id: ScArray,
+    address: ManagedAddress<M>,
+}
+
+#[derive(
+    TypeAbi, TopEncode, TopDecode, NestedEncode, NestedDecode, Clone, ManagedVecItem, PartialEq,
+)]
+pub enum ScArray {
+    None,
```
dorin-iancu, line 16: Do you ever need a 0-value, like `None`? If not, I'd say simply remove it and also remove the check.

chain-factory/src/factory.rs
```diff
@@ -36,6 +56,21 @@ pub trait FactoryModule {
         let _ = self.all_deployed_contracts().insert(sc_address);
     }
 
+    #[only_owner]
+    #[endpoint(addContractsToMap)]
+    fn add_contracts_to_map(
+        &self,
+        contracts_map: MultiValueEncoded<Self::Api, ContractMapArgs<Self::Api>>,
+    ) {
+        require!(!contracts_map.is_empty(), "Given contracts map is empty");
```
dorin-iancu, line 65: I'd say just remove this `require`. If it's empty, it does nothing anyway.

chain-factory/src/factory.rs
andreiblt1304: The problem is to avoid the 0-value because it won't be saved in the mapper

chain-factory/src/factory.rs
dorin-iancu: Actually, that's not true. An index of 0 will be nested encoded, so the storage key will be `contractsMap0`.

chain-factory/src/factory.rs
dorin-iancu: A 0 value is saved correctly in storage. Composed storage keys use nested encoding for members.

chain-factory/src/factory.rs
andreiblt1304: My bad, I remove the `None`

chain-factory/src/factory.rs
andreiblt1304: removed it

axenteoctavian (approved):

alyn509 (approved):

PR #110, opened 2024-08-20
Test fixes

andreiblt1304:


esdt-safe/tests/bridge_blackbox_tests.rs
```diff
@@ -278,6 +270,7 @@ impl BridgeTestState {
             .to(BRIDGE_ADDRESS)
             .typed(esdt_safe_proxy::EsdtSafeProxy)
             .execute_operations(hash_of_hashes, operation)
+            .returns(ExpectError(4, "Operation is not registered"))
```
axenteoctavian, line 273: Why is this returning error? 
You have this function that returns error `propose_execute_operation_and_expect_err`.

alyn509 (approved):

esdt-safe/tests/bridge_blackbox_tests.rs
alyn509: maybe just rename the test to `propose_execute_operation_fail` to be obvious that it expects fail..

andreiblt1304: I forgot about that function, and realized the error that was made, I fixed the test

PR #113, opened 2024-08-23
`Deposit` endpoint improvements

andreiblt1304:


enshrine-esdt-safe/src/to_sovereign/create_tx.rs
```diff
@@ -176,6 +157,18 @@ pub trait CreateTxModule:
         };
     }
 
+    fn require_gas_limit_under_limit(&self, gas_limit: &GasLimit) {
```
dorin-iancu, line 160: No need to take u64 as reference.

common/transaction/src/lib.rs
```diff
@@ -65,6 +65,35 @@ pub struct TransferData<M: ManagedTypeApi> {
...
+    ) -> Self {
+        TransferData {
+            gas_limit,
+            function,
+            args,
+        }
+    }
+
+    // TODO: Should use a reference
+    pub fn optional_value_to_option(
```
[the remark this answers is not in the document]
andreiblt1304: I'm not sure if it's worth to use a reference here since in the return from the `Some` branch will need a `clone()`

enshrine-esdt-safe/src/to_sovereign/create_tx.rs
andreiblt1304: removed

mihaicalinluca (approved):

alyn509 (approved):

PR #120, opened 2024-09-10
Enshrine ESDT unified syntax migration 

andreiblt1304:


enshrine-esdt-safe/src/to_sovereign/create_tx.rs
```diff
@@ -130,7 +130,7 @@ pub trait CreateTxModule:
         refundable_payments: &ManagedVec<EsdtTokenPayment>,
     ) {
         for payment in refundable_payments {
-            self.send().direct_non_zero_esdt_payment(caller, &payment);
+            self.tx().to(caller).payment(&payment).transfer();
```
axenteoctavian, line 133: You need a check if payment is non zero so you won't make transactions with 0 value. Or is the transfer function already checking this?

fee-market/src/subtract_fee.rs
```diff
@@ -123,14 +124,17 @@ pub trait SubtractFeeModule:
...
+
+        self.tokens_for_fees()
             .insert(final_payment.fee.token_identifier.clone());
+
         self.accumulated_fees(&final_payment.fee.token_identifier)
             .update(|amt| *amt += &final_payment.fee.amount);
 
-        self.send()
-            .direct_non_zero_esdt_payment(&original_caller, &final_payment.remaining_tokens);
+        self.tx()
```
axenteoctavian, line 134: You need a check if payment is non zero so you won't make transactions with 0 value. Or is the transfer function already checking this?

enshrine-esdt-safe/src/to_sovereign/create_tx.rs
andreiblt1304: added `require`

fee-market/src/subtract_fee.rs
andreiblt1304: added `require`

mihaicalinluca (approved):

PR #121, opened 2024-09-12
Liquid staking sc

andreiblt:


liquid-staking/src/liquidity_pools.rs
```diff
@@ -0,0 +1,4 @@
+use multiversx_sc::imports::*;
```
alyn509, line 1: delete this

liquid-staking/src/lib.rs
```diff
@@ -0,0 +1,27 @@
...
+
+#[multiversx_sc::contract]
+pub trait LiquidStaking:
+    liquidity_pools::LiquidityPoolModule
+    + delegation::DelegationModule
+    + common::storage::CommonStorageModule
+{
+    #[init]
+    fn init(&self, delegation_address: ManagedAddress) {
+        require!(
```
alyn509, line 17: better extract this into a function / endpoint `add_delegation_contract`. In the future you might have this contract working with more than 1 delegation contract.

liquid-staking/src/delegation.rs
```diff
@@ -0,0 +1,90 @@
...
+    ) {
+        match result {
+            ManagedAsyncCallResult::Ok(()) => {
+                self.delegated_value(caller.clone()).set(egld_amount);
+                self.egld_token_supply()
+                    .update(|value| *value += egld_amount)
+            }
+            _ => sc_panic!("There was an error at delegating"),
+        }
+    }
```
alyn509, line 70: just a small advice to ease code reading: put every callback after the endpoint that calls it

liquid-staking/src/delegation.rs
```diff
@@ -0,0 +1,90 @@
...
+            .call_and_exit();
+    }
+
+    #[endpoint(unStake)]
+    fn unstake(&self, egld_amount_to_unstake: BigUint) {
+        let caller = self.blockchain().get_caller();
+        let current_epoch = self.blockchain().get_block_epoch();
+        let total_egld_deposit = self.delegated_value(caller.clone()).get();
+        let delegation_contract_address = self.delegation_address().get();
+        let undelegate_endpoint = ManagedBuffer::from("unDelegate");
```
alyn509, line 30: make a pub const out of this

liquid-staking/src/delegation.rs
```diff
@@ -0,0 +1,90 @@
...
+
+        let mut args: ManagedArgBuffer<Self::Api> = ManagedArgBuffer::new();
+        args.push_arg(&egld_amount_to_unstake);
+
```
alyn509, line 34: I feel like it is way better to make a  check here if caller has something to unstake, by checking that the amount to unstake is available in `delegated_value`

in this way you will not call undelegate if you already know that the caller hasn't enough to unstake

liquid-staking/scenarios/liquid_staking.scen.json
```diff
@@ -0,0 +1,39 @@
...
+                }
+            ]
+        },
+        {
+            "step": "scDeploy",
+            "id": "deploy",
+            "tx": {
+                "from": "address:owner",
+                "contractCode": "mxsc:../output/liquid-staking.mxsc.json",
+                "arguments": [],
```
alyn509, line 26: you forgot the init args

alyn509 (changes requested):

liquid-staking/src/lib.rs
andreiblt1304: alright, in this case I will make the storage mapper to a `UnorderedSetMapper`

andreiblt1304: Nevermind, this won't work, I created a `SingleValueMapper` with an index as `contract_name`

alyn509 (approved):

delegation-mock/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+        );
+        self.address_deposit(&caller)
+            .update(|value| *value -= &egld_to_undelegate);
+        self.address_undelegate_amount(&caller)
+            .update(|value| *value += &egld_to_undelegate);
+        self.address_undelegate_epoch(&caller)
+            .set(current_epoch + UNBOND_PERIOD);
+    }
+
+    #[endpoint(withdraw)]
```
dorin-iancu, line 55: If the endpoint name is the same as the function name, you can simply use `#[endpoint]` annotation.

liquid-staking/src/common/storage.rs
```diff
@@ -0,0 +1,71 @@
...
+    #[view(getDelegationAddress)]
+    #[storage_mapper("delegationAddress")]
+    fn delegation_addresses(
+        &self,
+        contract_name: &ManagedBuffer,
+    ) -> SingleValueMapper<ManagedAddress>;
+
+    #[view(getDelegatedValue)]
+    #[storage_mapper("delegatedValue")]
+    fn delegated_value(&self, validator: &ManagedAddress) -> SingleValueMapper<BigUint<Self::Api>>;
```
dorin-iancu, line 15: No need to specify `<Self::Api>` for framework types.

```suggestion
    fn delegated_value(&self, validator: &ManagedAddress) -> SingleValueMapper<BigUint>;
```

liquid-staking/src/common/storage.rs
```diff
@@ -0,0 +1,71 @@
...
+    #[view(getRegisteredBlsKeys)]
+    #[storage_mapper("registeredBlsKeys")]
+    fn registered_bls_keys(&self) -> UnorderedSetMapper<ManagedBuffer>;
+
+    #[view(validatorBlsKeyMap)]
+    #[storage_mapper("validatorBlsKeyMap")]
+    fn validator_bls_key_address_map(
+        &self,
+        bls_key: &ManagedBuffer,
+    ) -> SingleValueMapper<ManagedAddress>;
```
dorin-iancu, line 38: Have a custom type for bls keys, i.e. `pub type BlsKey<M> = ManagedBuffer<M>;`

Makes the code a lot more readable, and it's very easy to change the type if you ever need to.

liquid-staking/src/common/storage.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    // NOTE: Number of nodes where ?
+    #[view(lockedSupply)]
+    #[storage_mapper("lockerSupply")]
+    fn locked_supply(&self, chain_id: ManagedBuffer) -> SingleValueMapper<BigUint<Self::Api>>;
```
dorin-iancu, line 43: Same thing for chain_id, `pub type ChainId<M> = ManagedBuffer<M>;`

liquid-staking/src/delegation.rs
```diff
@@ -0,0 +1,163 @@
+use multiversx_sc::imports::*;
+pub const UNBOND_PERIOD: u64 = 10;
+pub const DELEGATE_ENDPOINT: &str = "delegate";
+pub const UNDELEGATE_ENDPOINT: &str = "unDelegate";
+pub const CLAIM_REWARDS_ENDPOINT: &str = "claimRewards";
```
dorin-iancu, line 5: Use `&[u8]` instead of `&str`, you don't need all the overhead of strings. Also, add `b'` in front after you do.

dorin-iancu: Also, unsure why you insist on doing raw calls instead of defining a custom proxy. Example here:
https://github.com/multiversx/mx-ibc-rs/blob/some-tests/client-impls/local-host/src/client_logic.rs#L10-L24

Way less error-prone if you do it that way.

liquid-staking/src/delegation.rs
```diff
@@ -0,0 +1,163 @@
...
+    #[callback]
+    fn stake_callback(
+        &self,
+        caller: &ManagedAddress,
+        egld_amount: &BigUint,
+        #[call_result] result: ManagedAsyncCallResult<()>,
+    ) {
+        match result {
+            ManagedAsyncCallResult::Ok(()) => {
+                self.delegated_value(caller).set(egld_amount);
```
dorin-iancu, line 35: Shouldn't this `+=` instead of simply set? It would overwrite the value if the user stakes twice.

liquid-staking/src/delegation.rs
```diff
@@ -0,0 +1,163 @@
...
+                self.egld_token_supply()
+                    .update(|value| *value += egld_amount)
+            }
+            _ => sc_panic!("There was an error at delegating"),
```
dorin-iancu, line 39: I'm not sure if you should panic here. I think you should return the user's funds instead.

liquid-staking/src/delegation.rs
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+
+    #[endpoint(unStake)]
+    fn unstake(&self, contract_name: ManagedBuffer, egld_amount_to_unstake: BigUint) {
+        let caller = self.blockchain().get_caller();
+        self.require_caller_has_stake(&caller);
```
dorin-iancu, line 46: I'm not sure if you need this validation and the require below, as the delegation SC will throw an error anyway.

liquid-staking/src/delegation.rs
```diff
@@ -0,0 +1,163 @@
...
+        egld_amount_to_unstake: &BigUint,
+        current_epoch: Epoch,
+        #[call_result] result: ManagedAsyncCallResult<()>,
+    ) {
+        match result {
+            ManagedAsyncCallResult::Ok(()) => {
+                self.delegated_value(caller)
+                    .update(|value| *value -= egld_amount_to_unstake);
+                self.undelegate_epoch(caller)
+                    .set(current_epoch + UNBOND_PERIOD);
```
dorin-iancu, line 82: I'm not sure if you need this at all. First, it's overwritten whenever a user unstakes a second time. Second, just let the delegation SC handle all this unbond logic instead of duplicating it here.

liquid-staking/src/delegation.rs
```diff
@@ -0,0 +1,163 @@
...
+        for delegation_contract in contracts {
+            let delegation_mapper = self.delegation_addresses(&delegation_contract);
+            if !delegation_mapper.is_empty() {
+                let delegation_address = delegation_mapper.get();
+                self.tx()
+                    .to(delegation_address)
+                    .raw_call(CLAIM_REWARDS_ENDPOINT)
+                    .callback(DelegationModule::callbacks(self).claim_rewards_from_delegation_cb())
+                    .async_call_and_exit();
+            }
```
dorin-iancu, line 102: Slight de-nesting:

```suggestion
            if delegation_mapper.is_empty() {
                 continue;
            }
            
                let delegation_address = delegation_mapper.get();
                self.tx()
                    .to(delegation_address)
                    .raw_call(CLAIM_REWARDS_ENDPOINT)
                    .callback(DelegationModule::callbacks(self).claim_rewards_from_delegation_cb())
                    .async_call_and_exit();
            }
```

dorin-iancu: Btw, this won't work, since you're using the old async call, which means you're only allowed to do one per tx. Use the promise API instead.

liquid-staking/src/delegation.rs
```diff
@@ -0,0 +1,163 @@
...
+    #[callback]
+    fn claim_rewards_from_delegation_cb(
+        &self,
+        // egld_amount: &BigUint,
+        #[call_result] result: ManagedAsyncCallResult<()>,
+    ) {
+        match result {
+            ManagedAsyncCallResult::Ok(()) => {}
+            _ => sc_panic!("There was an error at claiming rewards from the delegation contract"),
+        }
```
dorin-iancu, line 115: If you don't need to do anything for `Ok()`:
```suggestion
        require!(result.is_ok(),  "There was an error at claiming rewards from the delegation contract");
```

liquid-staking/src/common/storage.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    fn require_bls_key_to_be_registered(&self, bls_key: &ManagedBuffer) {
```
dorin-iancu, line 45: Rename to `require_bls_key_registered`

liquid-staking/src/common/storage.rs
```diff
@@ -0,0 +1,71 @@
...
+        require!(
+            self.registered_bls_keys().contains(bls_key),
+            "The given bls key is not registered"
+        );
+    }
+
+    fn require_caller_to_be_header_verifier(&self, caller: &ManagedAddress) {
```
dorin-iancu, line 52: Rename to `require_caller_header_verifier`

liquid-staking/src/common/storage.rs
```diff
@@ -0,0 +1,71 @@
...
+
+        let header_verifier_address = self.header_verifier_address().get();
+
+        require!(
+            caller == &header_verifier_address,
+            "Caller is not Header Verifier contract"
+        );
+    }
+
+    fn require_caller_has_stake(&self, caller: &ManagedAddress) {
```
dorin-iancu, line 66: Rename to `require_address_has_stake`, and rename argument to `address: `

liquid-staking/src/delegation.rs
```diff
@@ -0,0 +1,163 @@
...
+        require!(value_to_slash > 0, "You can't slash a value of 0 eGLD");
+
+        let delegated_value = self.delegated_value(&validator_address).get();
+        require!(
+            delegated_value >= value_to_slash,
+            "The slash value can't be greater than the total delegated amount"
+        );
+
+        self.delegated_value(&validator_address)
+            .update(|value| *value -= &value_to_slash);
```
dorin-iancu, line 146: Slight optimization to only build the key once:

```suggestion
        let mapper = self.delegated_value(&validator_address).get();
        let delegated_value = mapper.get();
        require!(
            delegated_value >= value_to_slash,
            "The slash value can't be greater than the total delegated amount"
        );

        
           mapper.update(|value| *value -= &value_to_slash);
```

dorin-iancu: Also, consider using `AddressToIdMapper` where you have address as part of the key. Any storage key over 32 bytes costs extra gas to use.

dorin-iancu: Example: https://github.com/multiversx/sc-guilds-rs/blob/5b4f8eba503a173d2fbb6f110d33c79d4069e807/guild-factory/src/factory.rs#L191

And then, you simply have `fn delegated_value(&self, address_id: AddressId) -> SingleValueMapper<BigUint>;`

liquid-staking/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+    + delegation::DelegationModule
+    + common::storage::CommonStorageModule
+{
+    #[init]
+    fn init(&self) {}
+
+    #[endpoint(registerDelegationContractAddress)]
+    fn register_delegation_address(
+        &self,
+        contract_name: ManagedBuffer,
```
dorin-iancu, line 23: Unsure if you should allow custom names instead of assigning your internal IDs for contracts.

liquid-staking/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+
+        self.delegation_addresses(&contract_name)
+            .set(delegation_address);
+    }
+
+    #[only_owner]
+    #[endpoint(registerHeaderVerifierAddress)]
+    fn register_header_verifier_address(&self, header_verifier_address: ManagedAddress) {
+        self.header_verifier_address().set(header_verifier_address);
+    }
```
dorin-iancu, line 44: I think this could be simply on init function instead of requiring one additional call. You can keep the function in case you ever need to overwrite the address.

liquid-staking/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+
+    #[endpoint(registerBlsKeys)]
+    fn unregister_bls_keys(&self, bls_keys: MultiValueEncoded<ManagedBuffer>) {
+        let caller = self.blockchain().get_caller();
+        self.require_caller_to_be_header_verifier(&caller);
+
+        for bls_key in bls_keys {
+            self.registered_bls_keys().swap_remove(&bls_key);
+        }
```
dorin-iancu, line 61: Reuse the mapper instead of creating every single time:

```suggestion
        let mut mapper = self.registered_bls_keys();
        for bls_key in bls_keys {
            mapper.swap_remove(&bls_key);
        }
```

liquid-staking/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    #[upgrade]
+    fn upgrade(&self) {}
```
dorin-iancu, line 65: Just leave this right under `init`.

liquid-staking/tests/liquid_staking_tests.rs
```diff
@@ -0,0 +1,663 @@
...
+        );
+    }
+
+    fn get_expected_rewards(
+        &mut self,
+        staked_amount: &BigUint<StaticApi>,
+        current_epoch: u64,
+        last_claim_epoch: u64,
+    ) -> BigUint<StaticApi> {
+        let rewards = (staked_amount * APY / MAX_PERCENTAGE) * (current_epoch - last_claim_epoch)
```
dorin-iancu, line 299: Clippy :nutu:

liquid-staking/src/delegation.rs
andreiblt1304: Noted as TODO

liquid-staking/src/lib.rs
andreiblt1304: I was thinking about using an enum just like in `ChainFactory` contract

liquid-staking/src/lib.rs
andreiblt1304: I wanted to keep the init empty at the moment since there will probably be a lot more to put inside that

PR #126, opened 2024-09-17
Fixed failing esdt safe tests

andreiblt1304:


axenteoctavian (approved):
Unused functions but not important tests because this contract will be deprecated soon

PR #129, opened 2024-09-30
Header-Verifier Blackbox update + Coverage

andreiblt1304:
Also added another `require` for the `esdt_address` mapper so the error is more self explanatory

### DIFF
--- a/header-verifier/tests/header_verifier_blackbox_test.rs
+++ b/header-verifier/tests/header_verifier_blackbox_test.rs
@@ -1,125 +1,74 @@
-#![allow(deprecated)]
-
 use bls_signature::BlsSignature;
-use header_verifier::ProxyTrait;
-use multiversx_sc::types::{ManagedBuffer, ManagedByteArray, MultiValueEncoded};
+use header_verifier::header_verifier_proxy;
+use multiversx_sc::types::{
+    BigUint, ManagedBuffer, ManagedByteArray, ManagedVec, MultiValueEncoded, TestAddress,
+    TestSCAddress,
+};
 use multiversx_sc_scenario::{
     api::StaticApi,
-    scenario_model::{Account, ScCallStep, ScDeployStep, SetStateStep, TxExpect},
-    ContractInfo, ScenarioWorld,
+    imports::MxscPath,
+    scenario_model::{ScCallStep, ScDeployStep, TxExpect},
+    ContractInfo, ScenarioTxRun, ScenarioWorld,
 };
 
-const HEADER_VERIFIER_PATH_EXPR: &str = "file:output/header-verifier.wasm";
-const OWNER_ADDRESS_EXPR: &str = "address:owner";
-const LEADER_ADDRESS_EXPR: &str = "address:proposer";
-const VALIDATOR_ADDRESS_EXPR: &str = "address:board-member";
-const HEADER_VERIFIER_ADDRESS_EXPR: &str = "sc:header_verifier";
+const HEADER_VERIFIER_CODE_PATH: MxscPath = MxscPath::new("ouput/header-verifier.mxsc-json");
+const HEADER_VERIFIER_ADDRESS: TestSCAddress = TestSCAddress::new("header-verifier");
+const OWNER: TestAddress = TestAddress::new("owner");
+const LEADER: TestAddress = TestAddress::new("leader");
+const VALIDATOR: TestAddress = TestAddress::new("validator");
+const WEGLD_BALANCE: u128 = 100_000_000_000_000_000;
 
 type HeaderVerifierContract = ContractInfo<header_verifier::Proxy<StaticApi>>;
+type BlsKeys = MultiValueEncoded<StaticApi, ManagedBuffer<StaticApi>>;
 
 fn world() -> ScenarioWorld {
     let mut blockchain = ScenarioWorld::new();
-    blockchain.set_current_dir_from_workspace("header-verifier/src/lib");
-
-    blockchain.register_contract(HEADER_VERIFIER_PATH_EXPR, header_verifier::ContractBuilder);
+    blockchain.register_contract(HEADER_VERIFIER_CODE_PATH, header_verifier::ContractBuilder);
 
     blockchain
 }
 
 struct HeaderVerifierTestState {
     world: ScenarioWorld,
-    header_verifier_contract: HeaderVerifierContract,
 }
 
 impl HeaderVerifierTestState {
     fn new() -> Self {
         let mut world = world();
 
-        world.set_state_step(
-            SetStateStep::new()
-                .put_account(OWNER_ADDRESS_EXPR, Account::new().nonce(1))
-                .new_address(OWNER_ADDRESS_EXPR, 1, HEADER_VERIFIER_ADDRESS_EXPR)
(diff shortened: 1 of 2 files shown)


header-verifier/tests/header_verifier_blackbox_test.rs
```diff
@@ -1,125 +1,313 @@
...
+    MxscPath::new("../chain-factory/output/chain-factory.mxsc-json");
+const ENSHRINE_ADDRESS: TestSCAddress = TestSCAddress::new("enshrine");
+
+const OWNER: TestAddress = TestAddress::new("owner");
+const WEGLD_BALANCE: u128 = 100_000_000_000_000_000;
+
+type BlsKeys = MultiValueEncoded<StaticApi, ManagedBuffer<StaticApi>>;
 
-type HeaderVerifierContract = ContractInfo<header_verifier::Proxy<StaticApi>>;
+pub struct ErrorStatus<'a> {
```
mihaicalinluca, line 27: I think you can use ExpectError directly

andreiblt1304: i am using this struct because it's easier to match it inside a propose function, one arm is when an error is expected and one for no error

mihaicalinluca: an alternative would be to use Option<ExpectError> or otherwise ExpectError(0, "") represents success

andreiblt1304: each test that uses this struct has an Option<ExpectError>

alyn509 (approved):

mihaicalinluca (approved):

header-verifier/tests/header_verifier_blackbox_test.rs
```diff
@@ -1,125 +1,313 @@
...
+    state.propose_remove_execute_hash(&operation.bridge_operation_hash, operation_1, None);
+
+    state
+        .world
+        .query()
+        .to(HEADER_VERIFIER_ADDRESS)
+        .whitebox(header_verifier::contract_obj, |sc| {
+            let hash_of_hashes: ManagedBuffer<DebugApi> =
+                ManagedBuffer::from(operation.bridge_operation_hash.to_vec());
+            sc.pending_hashes(&hash_of_hashes).is_empty();
```
axenteoctavian, line 311: This doesn't look right. You register 2 operations, you remove 1 operation, but in the end expect that there are no pending_hashes and should be 1 left

header-verifier/tests/header_verifier_blackbox_test.rs
```diff
@@ -1,125 +1,313 @@
...
+    let operation_2 = ManagedBuffer::from("operation_2");
+    let operation = state.get_mock_operation(vec![&operation_1, &operation_2]);
+
+    let error_status = ErrorStatus {
+        code: 4,
+        error_message: "Only ESDT Safe contract can call this endpoint",
+    };
+
+    state.propose_register_operations(operation.clone());
+    state.propose_register_esdt_address(ENSHRINE_ADDRESS);
```
axenteoctavian, line 254: It is very hard to understand from this test that will return that error. You should add `caller` argument in `propose_remove_execute_hash` function

header-verifier/tests/header_verifier_blackbox_test.rs
```diff
@@ -1,125 +1,313 @@
...
+    state
+        .world
+        .query()
+        .to(HEADER_VERIFIER_ADDRESS)
+        .whitebox(header_verifier::contract_obj, |sc| {
+            let hash_of_hashes: ManagedBuffer<DebugApi> =
+                ManagedBuffer::from(operation.bridge_operation_hash.to_vec());
+            sc.pending_hashes(&hash_of_hashes).is_empty();
+        });
+}
```
axenteoctavian, line 313: Also please make another test that will remove all the operations that were registered and expect:
- pending_hashes is empty
- hashes history still contains that hash_of_hashes (this should not be cleared)

header-verifier/tests/header_verifier_blackbox_test.rs
```diff
@@ -1,125 +1,313 @@
...
-                .code(header_verifier_code),
+        bls_keys
+    }
+
+    fn get_mock_operation(
+        &mut self,
+        operations: Vec<&ManagedBuffer<StaticApi>>,
+    ) -> BridgeOperation<StaticApi> {
+        let mock_signature: BlsSignature<StaticApi> = ManagedByteArray::new_from_bytes(
+            b"EIZ2\x05\xf7q\xc7G\x96\x1f\xba0\xe2\xd1\xf5pE\x14\xd7?\xac\xff\x8d\x1a\x0c\x11\x900f5\xfb\xff4\x94\xb8@\xc5^\xc2,exn0\xe3\xf0\n"
```
axenteoctavian, line 151: This signature is not implemented at the moment. Just make it empty.

header-verifier/tests/header_verifier_blackbox_test.rs
```diff
@@ -1,125 +1,313 @@
...
-const HEADER_VERIFIER_PATH_EXPR: &str = "file:output/header-verifier.wasm";
-const OWNER_ADDRESS_EXPR: &str = "address:owner";
-const LEADER_ADDRESS_EXPR: &str = "address:proposer";
-const VALIDATOR_ADDRESS_EXPR: &str = "address:board-member";
-const HEADER_VERIFIER_ADDRESS_EXPR: &str = "sc:header_verifier";
+const HEADER_VERIFIER_CODE_PATH: MxscPath = MxscPath::new("ouput/header-verifier.mxsc-json");
+const HEADER_VERIFIER_ADDRESS: TestSCAddress = TestSCAddress::new("header-verifier");
+
+// NOTE: This is a mock path
+const ENSHRINE_CODE_PATH: MxscPath =
```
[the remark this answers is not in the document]
andreiblt1304: nope, you can't do that

header-verifier/tests/header_verifier_blackbox_test.rs
```diff
@@ -1,125 +1,313 @@
...
+            .typed(header_verifier_proxy::HeaderverifierProxy)
+            .init(bls_keys)
+            .code(HEADER_VERIFIER_CODE_PATH)
+            .new_address(HEADER_VERIFIER_ADDRESS)
+            .run();
+
+        self
+    }
+
+    fn deploy_enshrine_esdt_contract(&mut self, bls_keys: &BlsKeys) -> &mut Self {
```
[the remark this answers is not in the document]
andreiblt1304: you can't use just the address, you have to use a deployed smart contract

header-verifier/tests/header_verifier_blackbox_test.rs
andreiblt1304: I did that because test setup convenience, I will have to setup another contract for that. If i use it as an argument to the function I have to put it as `TestSCAddress`

header-verifier/tests/header_verifier_blackbox_test.rs
andreiblt1304: I just forgot the assert here, the test works as intended

PR #159, opened 2024-10-28
Chain-Factory modules cleaup

andreiblt1304:
Added separate crate for storage and util functions


axenteoctavian (approved):
If we'll decide to modify the chain_id logic, you can do it in another PR.

axenteoctavian (approved):

PR #166, opened 2024-11-05
Chain-Factory endpoints

andreiblt1304:
This contract will be deployed in each shard by the `master-chain-factory` SC. It's only purpose is to deploy the remaining contracts inside the Sovereign Chain.


alyn509 (approved):

chain-factory/src/factory.rs
```diff
@@ -160,7 +127,25 @@ pub trait FactoryModule:
         args.push_arg(price_aggregator_address);
 
         let fee_market_address = self.deploy_contract(source_address, args);
-        self.set_deployed_contract_to_storage(chain_id, ScArray::FeeMarket, fee_market_address);
+        self.set_deployed_contract_to_storage(chain_id, ScArray::FeeMarket, &fee_market_address);
+    }
+
+    #[only_owner]
+    #[endpoint(addContractsToMap)]
+    fn add_contracts_to_map(
```
axenteoctavian, line 135: This map is not used anywhere. I see you are using a storage for each template contract, not this `contract_mapper`.

chain-factory/tests/chain_factory_tests.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    blockchain
+}
+
+struct ChainFactoryTestState {
+    world: ScenarioWorld,
+}
+
+impl ChainFactoryTestState {
+    fn new(
```
axenteoctavian, line 39: This is `chain-factory` contract. Why on `new` you are deploying `chain-config` contract?

chain-factory/src/factory.rs
axenteoctavian, line 1: I see a lot of problems with this chain-factory contract implementation, like:
- after `deploy_sovereign_chain_config_contract` you can't even know what's the chain_id to use for deploying other contracts
- is not ok to make the caller an `admin`
- the "admin" can deploy contracts with whatever chain_id he wants
- and more...

A simpler way is to have a mapper with `key = caller` and value = (chain_id string, contracts ContractMapArgs).
So, when you deploy chain_config you create entry in mapper for caller, then for the rest of contracts you check mapper exists for caller and add the new contract address. This way you don't need `chain_id` in deploy function and everything is a lot simpler.

axenteoctavian, line 1: Just some notes for later updates:
- deploy contracts should be with proxy, not raw arguments
- factory contract should have a function that returns `true` if everything is ready for a chain to start, this needs to be used in esdt-safe to "complete setup phase"
- later, this contract should contain a new .rs file with a full deploy and setup phase and deposit+execute transactions

axenteoctavian (changes requested):

chain-factory/src/factory.rs
axenteoctavian: Other optimizations:
If you force user deploy in a custom order then you can add addresses in `init` to complete some mappers and you won't need extra transactions in setup phase
1. chain-config 
2. header-verifier
3. fee-market
4. enshrine-esdt-safe

- this order can also be optimized to deploy more contracts in the same transaction, like (1, 2-3, 4) or (1-2, 3-4) or other
- at step 4 you can also do:
1. add header-verifier and `fee-market` addresses on `init` in enshrine
2. add `enshrine` address in `header-verifier`
3. add whitelist for `enshrine` in `token-handler` 
this way we get rid of 4 extra transactions in setup phase

chain-factory/src/factory.rs
```diff
@@ -71,195 +38,90 @@ pub trait FactoryModule:
...
-    #[endpoint(deployCrossChainOperation)]
-    fn deploy_cross_chain_operation(
+    #[only_owner]
+    #[endpoint(deployEnshrineEsdtSafe)]
+    fn deploy_enshrine_esdt_safe(
         &self,
         is_sovereign_chain: bool,
+        token_handler_address: ManagedAddress,
         opt_wegld_identifier: Option<TokenIdentifier>,
         opt_sov_token_prefix: Option<ManagedBuffer>,
```
axenteoctavian, line 70: these should not be optional for main chain deploy
```suggestion
        wegld_identifier: TokenIdentifier,
        sov_token_prefix: ManagedBuffer,
```

chain-factory/src/factory.rs
```diff
@@ -1,64 +1,31 @@
...
-
 #[multiversx_sc::module]
-pub trait FactoryModule:
-    only_admin::OnlyAdminModule
-    + crate::common::storage::CommonStorage
-    + crate::common::utils::UtilsModule
-{
-    // TODO: Check if contract was already deployed
+pub trait FactoryModule {
     #[payable("EGLD")]
```
axenteoctavian, line 14: no need
```suggestion
```

axenteoctavian (approved):

PR #182, opened 2024-11-15
Add interactor tests for enshrine-esdt

sergiuosvat:


andreiblt1304:
you should use the new `common/proxies/` module instead of proxies inside each contract, it's far better and from this point on we will avoid as much as we can to have proxies in contract folders

mihaicalinluca (approved):

andreiblt1304 (approved):

PR #185, opened 2024-11-19
Sovereign-Forge SC

andreiblt1304:
Created the Sovereign-Forge SC that will call the Chain-Factory SC. Basically this contract will be the user's interaction with the Sovereign Chain in order to deploy and setup the needed modules and SCs. This PR contains the following:

- The newly generated Sovereign-Forge along side it's common modules: Utils and Storage
- The Phases module will call the endpoints from the Chain-Factory SC and deploy the needed contracts in `phases`


sovereign-forge/src/lib.rs
axenteoctavian, line 1: TODO for different PR
- define a "complete setup phase" for this contract where we check token-handlers and chain-factory storages are set
- only if the setup is completed, `deploy_phase_one` can be called

sovereign-forge/src/phases.rs
```diff
@@ -0,0 +1,84 @@
...
+        let call_value = self.call_value().egld_value();
+        self.require_correct_deploy_cost(call_value.deref());
+
+        let blockchain_api = self.blockchain();
+        let caller = blockchain_api.get_caller();
+        let caller_shard_id = blockchain_api.get_shard_of_address(&caller);
+
+        let chain_id = self.generate_chain_id();
+        let chain_factory_address = self.chain_factories(caller_shard_id).get();
+
```
axenteoctavian, line 38: require `chain_factory_address` is not empty

axenteoctavian (changes requested):

sovereign-forge/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+pub trait SovereignForge:
+    phases::PhasesModule + common::storage::StorageModule + common::utils::UtilsModule
+{
+    #[init]
+    fn init(&self, deploy_cost: BigUint) {
+        require!(deploy_cost > 0, "The deploy cost can't be a 0 value");
+        self.deploy_cost().set(deploy_cost);
+    }
+
+    #[endpoint(registerTokenHandler)]
```
axenteoctavian, line 19: Only owner can call this endpoint

sovereign-forge/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+
+        let blockchain_api = self.blockchain();
+        let caller = blockchain_api.get_caller();
+        let caller_shard_id = blockchain_api.get_shard_of_address(&caller);
+
+        self.token_handlers(caller_shard_id)
+            .set(token_handler_address);
+    }
+
+    #[endpoint(registerChainFactory)]
```
axenteoctavian, line 34: Only owner can call this endpoint

axenteoctavian (changes requested):

sovereign-forge/src/phases.rs
```diff
@@ -0,0 +1,92 @@
...
+        let caller = blockchain_api.get_caller();
+        let caller_shard_id = blockchain_api.get_shard_of_address(&caller);
+
+        let chain_id = self.generate_chain_id();
+
+        let chain_factories_mapper = self.chain_factories(caller_shard_id);
+        require!(
+            !chain_factories_mapper.is_empty(),
+            "There is no Chain-Factory address registered in the {} chain",
+            chain_id
```
axenteoctavian, line 42: Should be like this. I don't know if shard id of type u32 can be printed there
```suggestion
            "There is no Chain-Factory address registered in shard {}",
            caller_shard_id
```

axenteoctavian (approved):

mihaicalinluca (approved):

PR #192, opened 2024-11-26
Phase one

andreiblt1304:


chain-config/src/lib.rs
```diff
@@ -42,6 +41,44 @@ pub trait ChainConfigContract:
...
+        }
+    }
+
+    #[only_owner]
+    fn complete_setup_phase(&self) {
+        self.require_setup_not_complete();
+        self.require_config_set();
+        // validator set in header verifier
+        // change ownership to header-verifier
+        // update setup_phase_complete
```
axenteoctavian, line 79:
```suggestion
        self.setup_phase_complete().set(true);
```

chain-config/src/lib.rs
```diff
@@ -42,6 +41,44 @@ pub trait ChainConfigContract:
...
+
+                additional_stake_vec.push(value);
+            }
+            self.additional_stake_required().set(additional_stake_vec);
+        }
+    }
+
+    #[only_owner]
+    fn complete_setup_phase(&self) {
+        self.require_setup_not_complete();
```
axenteoctavian, line 75: This is not ok to fail, just use
`if is_setup_phase_complete() { return }`

PR #198, opened 2024-12-02
Interactor file structure refactor

sergiuosvat:


enshrine-esdt-safe/interactor/tests/enshrine-esdt-safe_interactor_tests.rs
```diff
@@ -0,0 +1,185 @@
...
+#[ignore]
+async fn test_deposit_too_many_payments() {
+    let mut interact = ContractInteract::new(Config::load_config()).await;
+    let to = interact.bob_address.clone();
+    let from = interact.wallet_address.clone();
+    let to_contract = interact.state.current_address().clone();
+    let transfer_data = OptionalTransferData::None;
+    let payments = ManagedVec::from(vec![
+        (
+            TokenIdentifier::from_esdt_bytes(TOKEN_ID),
```
andreiblt1304, line 59: Can you make a variable to populate this vector? Or maybe a getter function to return the whole vector? It is too much code in the test

andreiblt1304 (approved):

enshrine-esdt-safe/interactor/state.toml
axenteoctavian, line 1: Do you think this files should be ignored in `.gitignore`?
- Probably the addresses will change after every interactor run and you need to make sure to not include them in commits
- Maybe you don't want the addresses to be public so other people can interact with them

common/interactor/src/interactor_state.rs
```diff
@@ -0,0 +1,90 @@
...
+use std::{
+    io::{Read, Write},
+    path::Path,
+};
+
+const STATE_FILE: &str = "state.toml";
+
+#[derive(Debug, Default, Serialize, Deserialize)]
+pub struct State {
+    contract_address: Option<Bech32Address>,
```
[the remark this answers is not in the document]
sergiuosvat: Here the `contract_address` represents both `esdt_safe` and `enshrine_esdt_safe`, this file is imported in both interactors. Perhaps i should create separate variables to make the code cleaner

axenteoctavian (approved):

PR #224, opened 2025-01-28
Sovereign-Forge endpoint fixed

andreiblt1304:


sovereign-forge/src/phases.rs
```diff
@@ -39,18 +39,35 @@ pub trait PhasesModule:
             );
         }
 
+        let blockchain_api = self.blockchain();
+        let caller = blockchain_api.get_caller();
+        let caller_shard_id = blockchain_api.get_shard_of_address(&caller);
+
+        let chain_factory_address = self.chain_factories(caller_shard_id).get();
```
axenteoctavian, line 46: You can put all this in a function and use it here everywhere

axenteoctavian (approved):

PR #225, opened 2025-01-29
Operation register with setup phase

andreiblt1304:
Modified the operation registration logic in Header-Verifier to firstly check if the setup phase is completed. Also added test coverage and some cleanup.


enshrine-esdt-safe/tests/enshrine_esdt_safe_blackbox_test.rs
```diff
@@ -188,6 +192,32 @@ impl EnshrineTestState {
...
+                HEADER_VERIFIER_ADDRESS,
+            )
+            .code(CHAIN_CONFIG_CODE_PATH)
+            .new_address(CHAIN_CONFIG_ADDRESS)
+            .run();
+
+        self
+    }
+
+    fn complete_header_verifier_setup_phase(&mut self) {
```
axenteoctavian, line 211: esdt-safe does not call `complete_setup_phase` in header verifier

axenteoctavian (changes requested):

enshrine-esdt-safe/tests/enshrine_esdt_safe_blackbox_test.rs
andreiblt1304: `ENSHRINE_ESDT_OWNER_ADDRESS` is not a SC address

alyn509 (approved):

axenteoctavian (approved):

PR #234, opened 2025-02-18
DepositCommon module

andreiblt1304:


common/cross-chain/src/deposit_common.rs
```diff
@@ -0,0 +1,133 @@
...
+        let fee_payment = if fee_enabled_mapper {
+            OptionalValue::Some(self.pop_first_payment(payments.clone()).0)
+        } else {
+            OptionalValue::None
+        };
+
+        MultiValue2::from((fee_payment, payments))
+    }
+
+    fn burn_mainchain_token(
```
axenteoctavian, line 66: This function, some of below #[inline] functions and all mappers multiversx <-> sovereign should be only in `to-sovereign` contract, these are not common because they do not apply for enshine-esdt-safe

axenteoctavian (approved):
can be done in different PR

alyn509 (approved):

PR #248, opened 2025-03-01
ESDT-Safe optional config

andreiblt1304:
Modified in both the Sov-ESDT-Safe and Mvx-ESDT-Safe SC so that the `EsdtSafeConfig` parameter in `init` is now an `OptionalValue`. There is also a new endpoint in each SC so the config can be updated after deployment.

Also added the fee-market address inside the Sov-ESDT-Safe SC.

### DIFF
--- a/mvx-esdt-safe/src/lib.rs
+++ b/mvx-esdt-safe/src/lib.rs
@@ -24,11 +24,22 @@ pub trait MvxEsdtSafe:
     fn init(
         &self,
         header_verifier_address: ManagedAddress,
-        esdt_safe_config: EsdtSafeConfig<Self::Api>,
+        opt_config: OptionalValue<EsdtSafeConfig<Self::Api>>,
     ) {
         self.require_sc_address(&header_verifier_address);
         self.header_verifier_address().set(&header_verifier_address);
-        self.esdt_safe_config().set(esdt_safe_config);
+
+        self.esdt_safe_config().set(
+            opt_config
+                .into_option()
+                .unwrap_or_else(EsdtSafeConfig::default_config),
+        );
+    }
+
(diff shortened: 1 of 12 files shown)


mvx-esdt-safe/src/lib.rs
```diff
@@ -24,11 +24,22 @@ pub trait MvxEsdtSafe:
...
+            opt_config
+                .into_option()
+                .unwrap_or_else(EsdtSafeConfig::default_config),
+        );
+    }
+
+    #[only_owner]
+    #[endpoint(updateConfiguration)]
+    fn update_configuration(&self, new_config: EsdtSafeConfig<Self::Api>) {
+        self.esdt_safe_config().set(new_config);
```
axenteoctavian, line 42: Create a task for later to check that `new_config` is valid
1. max tx gas limit should be maximum `erd_max_gas_per_transaction` from [here](https://gateway.multiversx.com/network/config) - if you don't have access to this, hard-code the value

sov-esdt-safe/src/lib.rs
```diff
@@ -18,8 +18,25 @@ pub trait SovEsdtSafe:
...
+            opt_config
+                .into_option()
+                .unwrap_or_else(EsdtSafeConfig::default_config),
+        );
+    }
+
+    #[only_owner]
+    #[endpoint(updateConfiguration)]
+    fn update_configuration(&self, new_config: EsdtSafeConfig<Self::Api>) {
+        self.esdt_safe_config().set(new_config);
```
axenteoctavian, line 39: Same check that `new_config` is valid. Maybe we should have a common module for this if possible

sergiuosvat (approved):

PR #254, opened 2025-03-04
Add check esdt storage after deposit and esdt config unit tests

sergiuosvat:


alyn509 (approved):

mvx-esdt-safe/tests/mvx_esdt_safe_setup.rs
```diff
@@ -125,6 +125,16 @@ impl MvxEsdtSafeTestState {
         self
     }
 
+    pub fn update_configuration(&mut self, new_config: EsdtSafeConfig<StaticApi>) {
```
andreiblt1304, line 128: Can you please add here error case handling? I know that at the moment this endpoint won't throw any error, but it will in the future. Also add it in the sov-esdt-safe setup function as well

sov-esdt-safe/tests/sov_esdt_safe_setup.rs
```diff
@@ -241,4 +250,20 @@ impl SovEsdtSafeTestState {
             .returns(ReturnsLogs)
             .run()
     }
+
+    pub fn check_esdt_balance(&mut self, tokens: Vec<MultiValue3<TestTokenIdentifier, u64, u64>>) {
```
andreiblt1304, line 254: Rename to `check_sc_esdt_balance` and also add the SC address as parameter

andreiblt1304 (approved):

PR #255, opened 2025-03-04
Added parameters to interactor functions

andreiblt1304:


interactor/src/mvx_esdt_safe/mvx_esdt_safe_interactor_main.rs
```diff
@@ -114,82 +114,78 @@ impl MvxEsdtSafeInteract {
...
+    }
+
+    pub async fn register_token(
+        &mut self,
+        sov_token_id: &str,
+        token_type: EsdtTokenType,
+        token_display_name: &str,
+        token_ticker: &str,
+        num_decimals: u32,
+        egld_amount: BigUint<StaticApi>,
```
sergiuosvat, line 166: I think we can have a constant for egld_amount, unit tests should be enough for covering the wrong issue cost scenario

sergiuosvat (approved):

interactor/src/mvx_esdt_safe/mvx_esdt_safe_interactor_main.rs
andreiblt1304: it's better to have it as a parameter, we just use the constant when calling the function

interactor/src/mvx_esdt_safe/mvx_esdt_safe_interactor_main.rs
```diff
@@ -114,82 +114,78 @@ impl MvxEsdtSafeInteract {
...
+        to: Address,
+        opt_transfer_data: OptionalValueTransferDataTuple<StaticApi>,
+        payments: PaymentsVec<StaticApi>,
+    ) {
+        let response = self
+            .interactor
+            .tx()
+            .from(&self.wallet_address)
+            .to(self.state.current_address())
+            .gas(30_000_000u64)
```
alyn509, line 128: A good habit is to save numbers as constants, especially if you are using the same one multiple times.

Put it inside a constant names GAS_LIMIT or something

alyn509 (approved):

PR #264, opened 2025-03-05
Mvx-esdt-safe chain simulator tests

sergiuosvat:


interactor/src/mvx_esdt_safe/mvx_esdt_safe_interactor_main.rs
```diff
@@ -22,59 +32,261 @@ impl MvxEsdtSafeInteract {
...
         // generate blocks until ESDTSystemSCAddress is enabled
-        interactor.generate_blocks_until_epoch(1).await.unwrap();
+        interactor.generate_blocks(30u64).await.unwrap();
+
+        let set_state_response = interactor.set_state_for_saved_accounts().await;
+        interactor.generate_blocks(2u64).await.unwrap();
+        assert!(set_state_response.is_ok());
+
+        let mvx_esdt_safe_contract_code = BytesValue::interpret_from(
+            "mxsc:../../../mvx-esdt-safe/output/mvx-esdt-safe.mxsc.json",
```
andreiblt1304, line 47: Please create constants with the sc code paths

interactor/src/mvx_esdt_safe/mvx_esdt_safe_interactor_main.rs
```diff
@@ -1,18 +1,28 @@
...
 pub struct MvxEsdtSafeInteract {
-    interactor: Interactor,
-    wallet_address: Address,
-    contract_code: BytesValue,
-    state: State,
+    pub interactor: Interactor,
+    pub wallet_address: Address,
+    pub bob_address: Address,
+    mvx_esdt_safe_contract_code: BytesValue,
+    header_verifier_mvx_esdt_safe_contract_code: BytesValue,
```
andreiblt1304, line 21: i think it's enough to call it `header_verifier_contract_code`

interactor/tests/mvx_esdt_safe_tests.rs
```diff
@@ -0,0 +1,1258 @@
...
+                .clone(),
+            OptionalValue::Some(config),
+        )
+        .await;
+
+    chain_interactor
+        .interactor
+        .generate_blocks(2u64)
+        .await
+        .unwrap();
```
andreiblt1304, line 391: Why do you need to `unrwap()` after awaiting this call ?

sergiuosvat: `generate_blocks` returns whether it was successful or not and i do not want to retrieve that value. However, these calls will be removed in the next commit

interactor/src/interact.rs
```diff
@@ -14,6 +13,30 @@ use std::{
...
 const STATE_FILE: &str = "state.toml";
+pub const FEE_TOKEN: &str = "INTERNS-eaad15";
+pub const FIRST_TOKEN: &str = "GREEN-0e161c";
+pub const TOKEN_TICKER: &str = "GREEN";
+pub const SOV_TOKEN: &str = "sov-GREEN-0e161c";
+pub const SECOND_TOKEN: &str = "LTST-4f849e";
+pub const ISSUE_COST: u64 = 50_000_000_000_000_000; // 0.05 EGLD
+pub const SOV_TO_MVX_TOKEN_STORAGE_KEY: &str = "sovToMxTokenId";
+pub const MVX_TO_SOV_TOKEN_STORAGE_KEY: &str = "mxToSovTokenId";
+pub const MVX_ESDT_SAFE_CONTRACT_CODE: &str =
```
andreiblt1304, line 24: Can we use the constants from the common blackbox module?

interactor/src/interact.rs
```diff
@@ -14,6 +13,30 @@ use std::{
...
+    "mxsc:../../../mvx-esdt-safe/output/mvx-esdt-safe.mxsc.json";
+pub const HEADER_VERIFIER_CONTRACT_CODE: &str =
+    "mxsc:../../../header-verifier/output/header-verifier.mxsc.json";
+pub const FEE_MARKET_CONTRACT_CODE: &str = "mxsc:../../../fee-market/output/fee-market.mxsc.json";
+pub const TESTING_SC_CONTRACT_CODE: &str = "mxsc:../../../testing-sc/output/testing-sc.mxsc.json";
+pub const CHAIN_CONFIG_SC_CONTRACT_CODE: &str =
+    "mxsc:../../../chain-config/output/chain-config.mxsc.json";
+
+pub struct RegisterTokenArgs<'a> {
```
andreiblt1304, line 33: Can't you use the struct that is already declared in the mvx-esdt-safe blackbox module ? Or even move it in the common setup module

interactor/tests/mvx_esdt_safe_tests.rs
```diff
@@ -0,0 +1,1291 @@
...
+use rust_interact::mvx_esdt_safe::mvx_esdt_safe_interactor_main::MvxEsdtSafeInteract;
+use rust_interact::{
+    RegisterTokenArgs, FEE_TOKEN, FIRST_TOKEN, ISSUE_COST, MVX_TO_SOV_TOKEN_STORAGE_KEY,
+    SECOND_TOKEN, SOV_TOKEN, SOV_TO_MVX_TOKEN_STORAGE_KEY, TOKEN_TICKER,
+};
+use serial_test::serial;
+use structs::aliases::PaymentsVec;
+use structs::configs::EsdtSafeConfig;
+use structs::operation::{Operation, OperationData, OperationEsdtPayment, TransferData};
+
```
andreiblt1304, line 15: Would be nice to also add the steps for those tests as well, just like the steps for the blackbox tests

interactor/tests/mvx_esdt_safe_tests.rs
```diff
@@ -0,0 +1,1291 @@
...
+    let response = chain_interactor
+        .interactor
+        .set_state_overwrite(state_vec)
+        .await;
+    chain_interactor
+        .interactor
+        .generate_blocks(2u64)
+        .await
+        .unwrap();
+    assert!(response.is_ok());
```
andreiblt1304, line 154: Is there any way to add these calls to a function so it's less duplicated code ?

interactor/tests/mvx_esdt_safe_tests.rs
```diff
@@ -0,0 +1,1291 @@
...
+}
+
+#[tokio::test]
+#[serial]
+#[cfg_attr(not(feature = "chain-simulator-tests"), ignore)]
+async fn deposit_no_transfer_data() {
```
andreiblt1304, line 160: Please specify if the fee is enabled or not

interactor/tests/mvx_esdt_safe_tests.rs
```diff
@@ -0,0 +1,1291 @@
...
+        .generate_blocks(2u64)
+        .await
+        .unwrap();
+    assert!(response.is_ok());
+}
+
+#[tokio::test]
+#[serial]
+#[cfg_attr(not(feature = "chain-simulator-tests"), ignore)]
+async fn deposit_payment_doesnt_cover_fee() {
```
andreiblt1304, line 600: `deposit_payment_does_not_cover_fee`

interactor/tests/mvx_esdt_safe_tests.rs
```diff
@@ -0,0 +1,1291 @@
...
+        .await;
+
+    chain_interactor
+        .check_account_storage(
+            chain_interactor
+                .state
+                .current_mvx_esdt_safe_contract_address()
+                .clone()
+                .to_address(),
+            MVX_TO_SOV_TOKEN_STORAGE_KEY,
```
andreiblt1304, line 246: I'm not sure why do you check for this here

interactor/tests/mvx_esdt_safe_tests.rs
```diff
@@ -0,0 +1,1291 @@
...
+        .generate_blocks(2u64)
+        .await
+        .unwrap();
+    assert!(response.is_ok());
+}
+
+#[tokio::test]
+#[serial]
+#[cfg_attr(not(feature = "chain-simulator-tests"), ignore)]
+async fn register_token_non_fungible() {
```
andreiblt1304, line 835: I think it's easier to read to have this naming convention for the `register_token` tests:

`register_token_*token_type*_token`

interactor/tests/mvx_esdt_safe_tests.rs
```diff
@@ -0,0 +1,1291 @@
...
+
+    let operations_hashes = MultiValueEncoded::from(ManagedVec::from(vec![operation_hash.clone()]));
+
+    chain_interactor.deploy_chain_config().await;
+
+    chain_interactor
+        .register_operation(ManagedBuffer::new(), &hash_of_hashes, operations_hashes)
+        .await;
+
+    chain_interactor
```
andreiblt1304, line 1135: I'm not sure if you do any checks here after the `execute` endpoint, is there an implementation issue ?

testing-sc/Cargo.toml
andreiblt1304, line 1: Revert this

interactor/tests/mvx_esdt_safe_tests.rs
andreiblt1304, line 1: Please use the common constant error messages for response assertions

interactor/src/mvx_esdt_safe/mvx_esdt_safe_interactor_main.rs
```diff
@@ -21,60 +35,288 @@ impl MvxEsdtSafeInteract {
...
+            .tx()
+            .from(&self.wallet_address)
+            .gas(60_000_000u64)
+            .typed(FeeMarketProxy)
+            .init(esdt_safe_address, fee)
+            .code(&self.fee_market_contract_code)
+            .code_metadata(CodeMetadata::all())
+            .returns(ReturnsNewAddress)
+            .run()
+            .await;
```
andreiblt1304, line 211: Add an empty line after each deploy and adding the address to the storage, it is easier to read

interactor/src/mvx_esdt_safe/mvx_esdt_safe_interactor_main.rs
```diff
@@ -120,80 +362,103 @@ impl MvxEsdtSafeInteract {
...
             .typed(MvxEsdtSafeProxy)
             .deposit(to, opt_transfer_data)
             .payment(payments)
-            .returns(ReturnsResultUnmanaged)
+            .returns(ReturnsHandledOrError::new())
             .run()
             .await;
 
-        println!("Result: {response:?}");
+        match response {
```
andreiblt1304, line 380: I think you can use the common blackbox setup function for handling the error message

interactor/src/mvx_esdt_safe/mvx_esdt_safe_interactor_main.rs
```diff
@@ -268,4 +533,15 @@ impl MvxEsdtSafeInteract {
 
         println!("Result: {result_value:?}");
     }
+
+    pub fn get_operation_hash(
```
andreiblt1304, line 537: Put as a `TODO`, i see there are multiple `get_operation_hash` functions in the Sovereign repo, i think it should be added as a common function

mihaicalinluca (approved):

interactor/src/mvx_esdt_safe/mvx_esdt_safe_interactor_main.rs
```diff
@@ -21,60 +34,276 @@ impl MvxEsdtSafeInteract {
...
+        wanted_key: &str,
+        expected_value: Option<&str>,
+    ) {
+        let pairs = self.interactor.get_account_storage(&address).await;
+
+        let wanted_hex = hex::encode(wanted_key);
+
+        let found_entry = pairs.iter().find(|(key, _)| key.contains(&wanted_hex));
+
+        match (found_entry, expected_value) {
```
axenteoctavian, line 102: - `(Some(_), None)` can be a valid case, you can just find a key and not check its value (expected_value is Option)
- `(None, None)` should not be a valid case

I think it's too complicated this `match`. My suggestion:
- assert found_entry is_some, "Expected key containing '{}' was not found in account storage." -> with this you make sure the key is in account because that's why you call the function
- if expected_value is_some => assert found_entry value == expected_value -> with this you make sure the value is correct, if sent in the arguments.

interactor/src/interact.rs
```diff
@@ -14,6 +13,13 @@ use std::{
 };
 
 const STATE_FILE: &str = "state.toml";
+pub const FEE_TOKEN: &str = "INTERNS-eaad15";
+pub const FIRST_TOKEN: &str = "GREEN-0e161c";
+pub const TOKEN_TICKER: &str = "GREEN";
+pub const SOV_TOKEN: &str = "sov-GREEN-0e161c";
+pub const SECOND_TOKEN: &str = "LTST-4f849e";
+pub const MVX_ESDT_SAFE_CONTRACT_PATH: MxscPath =
```
axenteoctavian, line 21: Why this const is not in common blackbox setup file?

interactor/src/mvx_esdt_safe/mvx_esdt_safe_interactor_main.rs
```diff
@@ -268,4 +508,46 @@ impl MvxEsdtSafeInteract {
...
+        operation: &Operation<StaticApi>,
+    ) -> ManagedBuffer<StaticApi> {
+        let mut serialized_operation: ManagedBuffer<StaticApi> = ManagedBuffer::new();
+        let _ = operation.top_encode(&mut serialized_operation);
+        let sha256 = sha256(&serialized_operation.to_vec());
+
+        ManagedBuffer::new_from_bytes(&sha256)
+    }
+
+    pub async fn reset_state_chain_sim(&mut self, address_states: Option<Vec<Bech32Address>>) {
```
axenteoctavian, line 524: Question: What exactly is this function doing and why you need it?
Isn't `let mut chain_interactor = MvxEsdtSafeInteract::new(Config::chain_simulator_config()).await;` starting a fresh chain simulator for each test?

interactor/tests/mvx_esdt_safe_tests.rs
```diff
@@ -0,0 +1,1391 @@
...
+// 3. Deploy fee-market smart contract
+// 4. Set fee-market address in mvx-esdt-safe smart contract
+// 5. Unpause mvx-esdt-safe smart contract
+// 6. Deposit tokens
+#[tokio::test]
+#[serial]
+#[cfg_attr(not(feature = "chain-simulator-tests"), ignore)]
+async fn deposit_nothing_to_transfer_no_fee() {
+    let mut chain_interactor = MvxEsdtSafeInteract::new(Config::chain_simulator_config()).await;
+    chain_interactor.deploy_header_verifier().await;
```
axenteoctavian, line 33: Please create a function to deploy all these contacts. You can send EsdtConfig, FeeConfig as arguments
- deploy_header_verifier
- deploy_mvx_esdt_safe
- deploy_fee_market
- set_fee_market_address
- unpause_endpoint

axenteoctavian: If you don't need all these deploys in some test, just don't call the function and deploy manually

interactor/tests/mvx_esdt_safe_tests.rs
axenteoctavian, line 1: I see probably all these tests are duplicated with the blackbox tests. This is not really good practice. Also, I don't think it's necessary to run the same test with both blackbox and chain simulator.
We'll need to define some specific and more complex tests for this interactor.

andreiblt1304: In my opinion it is the best way to coverage the smart contracts. Blackbox tests are easier to write but might have some different outputs since, at the moment, the RustVM is not an 1:1 representation of the GoVM on the Devnet. I think all the blackbox tests should be duplicated within the interactor. With that we are 100% sure the flow works as intended and, as a plus, we test the RustVM. 

Related to your point, I agree, the interactor should have more complex scenarios, but at the moment this is what we have. @sergiuosvat has been working on a testing coverage spreadsheet and we can take up the needed scenarios over there.

interactor/src/mvx_esdt_safe/mvx_esdt_safe_interactor_main.rs
sergiuosvat: There was some kind of a racing issue when running multiple tests. Let's say Test A checks that deposit works properly (to do this we have to deploy header verifier etc) then we do another Test B that checks again some particular feature of the deposit. The issue was with the states, when Test A finished, the changes made inside the test were not reset. Thus, when Test B comes, the chain sim fails when trying to deploy the contracts again. That function solves this problem: erases everything that was done in the test so that we can truly have a fresh instance at the start of the next test

andreiblt1304 (approved):

interactor/src/mvx_esdt_safe/mvx_esdt_safe_interactor_main.rs
```diff
@@ -95,30 +94,49 @@ impl MvxEsdtSafeInteract {
...
         let pairs = self.interactor.get_account_storage(&address).await;
 
-        let wanted_hex = hex::encode(wanted_key);
+        let found_entry = pairs.iter().find(|(key, _)| key.contains(wanted_key));
 
-        let found_entry = pairs.iter().find(|(key, _)| key.contains(&wanted_hex));
+        let decoded_key = hex::decode(wanted_key)
+            .ok()
+            .and_then(|bytes| String::from_utf8(bytes).ok())
+            .unwrap_or_else(|| "<invalid utf-8>".to_string());
```
axenteoctavian, line 102: If the `wanted_key` cannot be decoded, then `decoded_key` will have value `<invalid utf-8>` or the test will fail?

I would suggest to create a new function for this decoding (and all from below). Also use some assert for decoded to not have value `<invalid utf-8>`

interactor/src/mvx_esdt_safe/mvx_esdt_safe_interactor_main.rs
axenteoctavian: I see you call this function sometimes with some arguments. Maybe it would be easier to erase everything all the time not only some addresses based on each test, it's your choice... :smile:

interactor/tests/mvx_esdt_safe_tests.rs
axenteoctavian: All the definition of payments and all the stuff is duplicated and will be harder to manage and prone to errors.

interactor/src/mvx_esdt_safe/mvx_esdt_safe_interactor_main.rs
sergiuosvat: At first, i tried the option of having a global reset state function but if some tests do not deploy all smart contracts present in the reset state, the transactions in those tests would end up with error: `Transaction still processing; Cannot generate blocks`. Therefore, i had to come up with a workaround. In the future, i hope these issues will be solved and we won't have to rely on workarounds...

interactor/tests/mvx_esdt_safe_tests.rs
sergiuosvat: There will be a follow up PR in which we clean everyting up so that we use only the common setup for payments and so on across all types of tests

axenteoctavian (approved):

PR #297, opened 2025-03-28
Tests cleanup

andreiblt1304:


common/common-blackbox-setup/src/lib.rs
```diff
@@ -203,6 +204,18 @@ impl BaseSetup {
             });
     }
 
+    pub fn check_deposited_tokens_amount_is_empty(&mut self, tokens: Vec<TestTokenIdentifier>) {
```
axenteoctavian, line 207: Empty is just if some token was never deposited.
I would say you should check for some value (can either be zero or some other value).
After deposit has some value, then after execute should be zero.

mvx-esdt-safe/tests/mvx_esdt_safe_blackbox_unit_tests.rs
axenteoctavian, line 1: GitHub warning with unused import in this file

axenteoctavian (approved):

sergiuosvat (approved):

PR #298, opened 2025-03-31
TransferData only endpoint calls

andreiblt1304:
This feature would allow the ESDT-Safe smart contracts to execute other smart contract calls without the need of payments.

### DIFF
--- a/mvx-esdt-safe/src/deposit.rs
+++ b/mvx-esdt-safe/src/deposit.rs
@@ -47,7 +47,7 @@ pub trait DepositModule:
         let (fees_payment, payments) = self.check_and_extract_fee().into_tuple();
         let mut total_tokens_for_fees = 0usize;
         let mut event_payments = MultiValueEncoded::new();
-        let mut refundable_payments = ManagedVec::<Self::Api, _>::new();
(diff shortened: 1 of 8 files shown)


mvx-esdt-safe/src/deposit.rs
```diff
@@ -23,11 +23,31 @@ pub trait DepositModule:
...
+
+        if let Some(transfer_data) = &option_transfer_data {
+            self.require_gas_limit_under_limit(transfer_data.gas_limit);
+            self.require_endpoint_not_banned(&transfer_data.function);
+        }
+
+        let has_no_esdt_transfers = self.call_value().all_esdt_transfers().clone().is_empty();
+        let deposit_without_transfers = has_no_esdt_transfers && opt_transfer_data.is_some();
+
+        if deposit_without_transfers {
```
CostinCarabas, line 36: isn't `!has_no_esdt_transfers` enough for a check?

andreiblt1304: no, if the `opt_transfer_data` variable is `None` that means that the endpoint won't do any smart contract calls so it doesn't need to call the new function

PR #314, opened 2025-04-08
Logs assertion overhaul

andreiblt1304:
This PR contains improvements for the blackbox common setup module and also how the logs are being handled.


sergiuosvat (approved):

common/common-blackbox-setup/src/lib.rs
```diff
@@ -270,4 +265,46 @@ impl BaseSetup {
             },
         )
     }
+
+    pub fn handle_endpoint_response_and_logs(
```
axenteoctavian, line 269: I would split this function to handle 1) error 2) logs. You could check logs even if the tx is failed (for example execute to check confirmation+deposit back).

mvx-esdt-safe/tests/mvx_esdt_safe_blackbox_setup.rs
```diff
@@ -241,37 +270,12 @@ impl MvxEsdtSafeTestState {
         &mut self,
         to: ManagedAddress<StaticApi>,
         opt_transfer_data: OptionalValueTransferDataTuple<StaticApi>,
-        opt_payment: Option<PaymentsVec<StaticApi>>,
+        payment: PaymentsVec<StaticApi>,
         expected_error_message: Option<&str>,
-    ) {
-        let tx = self
+        expected_log: Option<&str>,
+    ) -> Option<Log> {
```
axenteoctavian, line 276: I don't think you need to return `Log`. I see you always check that log is not empty which is probably always true because (I believe) there will be logs for both successful and failed txs.

axenteoctavian, line 276: Also, you check expected logs and also return them which looks kind of bad.
In my opinion, the best option would be to return (logs, response) from deposit, then inside the test you validate response success/fail with expected error, then you can check multiple logs (deposit, confirmation, also mint/burn or whatever)... but this involves a bigger refactor

header-verifier/tests/header_verifier_blackbox_setup.rs
```diff
@@ -1,19 +1,16 @@
...
+    api::StaticApi, multiversx_chain_vm::crypto_functions::sha256, ScenarioTxRun,
 };
 use multiversx_sc_scenario::{ReturnsHandledOrError, ReturnsLogs};
 use proxies::header_verifier_proxy::HeaderverifierProxy;
 
-const HEADER_VERIFIER_CODE_PATH: MxscPath = MxscPath::new("ouput/header-verifier.mxsc-json");
-pub const HEADER_VERIFIER_ADDRESS: TestSCAddress = TestSCAddress::new("header-verifier");
-
 // NOTE: This is a mock path
 pub const ENSHRINE_ADDRESS: TestAddress = TestAddress::new("enshrine");
```
axenteoctavian, line 15: I guess you could use the constants from common blackbox setup for all of these.

mvx-esdt-safe/tests/mvx_esdt_safe_blackbox_setup.rs
andreiblt1304: I think handling inside the test would be a bad practice, one good improvement would be to pass to the handle function an array of expected logs

sergiuosvat (approved):

axenteoctavian (approved):

PR #337, opened 2025-05-05
Max amount in esdt safe config

andreiblt1304:


common/cross-chain/src/deposit_common.rs
```diff
@@ -95,12 +95,11 @@ pub trait DepositCommonModule:
...
-        if max_amount > 0 {
-            amount > &max_amount
-        } else {
-            false
-        }
+        self.esdt_safe_config()
+            .get()
+            .max_bridged_token_amounts
+            .iter()
+            .any(|m| m.token_id == *token_id && amount <= &m.amount)
```
axenteoctavian, line 102: Your amount check is different than the previous version.
1. Here is the ChatGPT suggestion
```suggestion
    self.esdt_safe_config()
        .get()
        .max_bridged_token_amounts
        .iter()
        .find(|m| m.token_id == *token_id)
        .map(|m| amount > &m.amount)
        .unwrap_or(false)
```
2. Please add some unit tests

axenteoctavian (changes requested):

axenteoctavian (approved):

alyn509 (approved):

PR #339, opened 2025-05-07
Update mvx-esdt-safe and sovereign config through sovereign-forge

andreiblt1304:


alyn509 (approved):

sovereign-forge/tests/sovereign_forge_unit_tests.rs
```diff
@@ -354,6 +418,192 @@ fn register_chain_factory() {
...
+            assert!(!sc.chain_factories(2).is_empty());
+        });
+
+    let deploy_cost = BigUint::from(100_000u32);
+
+    state.complete_setup_phase(None);
+
+    state.deploy_phase_one(
+        &deploy_cost,
+        Some(ManagedBuffer::from("SVCH")),
```
axenteoctavian, line 465: This is prefered_chain_id but it needs to be 1-4 alphanumeric lowercase characters. We need to check this in `generate_chain_id` function

axenteoctavian (approved):

PR #352, opened 2025-05-16
Clean-up unit tests

sergiuosvat:


enshrine-esdt-safe/tests/enshrine_esdt_safe_blackbox_setup.rs
```diff
@@ -174,15 +173,24 @@ impl EnshrineTestState {
             .typed(EnshrineEsdtSafeProxy)
             .execute_operations(hash_of_hashes, operation)
             .returns(ReturnsHandledOrError::new())
+            .returns(ReturnsLogs)
             .run();
 
+        print!("Logs: {:?}", logs);
```
andreiblt1304, line 179: I don't think we need the here, let's keep printing just for the interactor

sergiuosvat: I did some debugging and forgot to remove it

CostinCarabas (approved):

PR #355, opened 2025-05-20
Add small test improvements

sergiuosvat:


common/common-test-setup/src/lib.rs
```diff
@@ -52,6 +53,11 @@ pub struct AccountSetup<'a> {
     pub egld_balance: Option<BigUint<StaticApi>>,
 }
 
+pub enum CallerAddress {
+    Owner,
+    Enshrine,
```
andreiblt1304, line 58: This enum is a great idea! But for ease of use and understanding let's rename `Enshrine` to `SafeSC` since it doesn't matter which ESDT contract we use

andreiblt1304 (approved):

PR #356, opened 2025-05-20
Added config update in Sovereign-Forge

andreiblt1304:


sovereign-forge/tests/sovereign_forge_blackbox_tests.rs
```diff
@@ -329,6 +334,194 @@ fn test_update_esdt_safe_config() {
 ///
 /// ### EXPECTED
 /// Setup phase is complete
```
sergiuosvat, line 336: The documentation does not match the test here

PR #359, opened 2025-05-22
Hash locking for config update endpoints

andreiblt1304:
Since the configuration updates can be run after the setup-phase completion, the hash of the operation must be registered then locked then removed.

Also added a new trait, `GenerateHash`, that implements a `generate_hash` method.

### DIFF
--- a/chain-config/src/lib.rs
+++ b/chain-config/src/lib.rs
@@ -1,6 +1,7 @@
 #![no_std]
 
-use structs::configs::SovereignConfig;
+use cross_chain::{execute_common, storage};
+use structs::{configs::SovereignConfig, generate_hash::GenerateHash};
 
 multiversx_sc::imports!();
 
@@ -8,7 +9,10 @@ pub mod validator_rules;
 
 #[multiversx_sc::contract]
 pub trait ChainConfigContract:
-    validator_rules::ValidatorRulesModule + setup_phase::SetupPhaseModule
+    validator_rules::ValidatorRulesModule
+    + setup_phase::SetupPhaseModule
+    + execute_common::ExecuteCommonModule
+    + storage::CrossChainStorage
 {
     #[init]
     fn init(&self, config: SovereignConfig<Self::Api>) {
@@ -18,9 +22,23 @@ pub trait ChainConfigContract:
 
     #[only_owner]
     #[endpoint(updateConfig)]
-    fn update_config(&self, new_config: SovereignConfig<Self::Api>) {
+    fn update_config(&self, hash_of_hashes: ManagedBuffer, new_config: SovereignConfig<Self::Api>) {
+        let opt_hash = if self.is_setup_phase_complete() {
+            Some(new_config.generate_hash())
+        } else {
+            None
+        };
+
+        if let Some(ref config_hash) = opt_hash {
+            self.lock_operation_hash(config_hash, &hash_of_hashes);
+        }
+
         self.require_valid_config(&new_config);
         self.sovereign_config().set(new_config);
+
+        if let Some(config_hash) = opt_hash {
+            self.remove_executed_hash(&hash_of_hashes, &config_hash);
+        }
     }
 
     #[only_owner]
--- a/chain-config/src/lib.rs
+++ b/chain-config/src/lib.rs
@@ -36,7 +36,7 @@ pub trait ChainConfigContract:
         self.lock_operation_hash(&config_hash, &hash_of_hashes);
 
         if let Some(error_message) = self.is_new_config_valid(&new_config) {
-            self.update_sovereign_config_fail_event(
+            self.failed_bridge_operation_event(
                 &hash_of_hashes,
                 &config_hash,
                 &ManagedBuffer::from(error_message),
--- a/chain-config/src/lib.rs
+++ b/chain-config/src/lib.rs
@@ -1,6 +1,6 @@
 #![no_std]
 
-use cross_chain::{execute_common, storage};
+use cross_chain::events;
 use structs::{configs::SovereignConfig, generate_hash::GenerateHash};
 
 multiversx_sc::imports!();
@@ -9,10 +9,7 @@ pub mod validator_rules;
 
 #[multiversx_sc::contract]
 pub trait ChainConfigContract:
-    validator_rules::ValidatorRulesModule
-    + setup_phase::SetupPhaseModule
-    + execute_common::ExecuteCommonModule
-    + storage::CrossChainStorage
+    validator_rules::ValidatorRulesModule + setup_phase::SetupPhaseModule + events::EventsModule
 {
     #[init]
     fn init(&self, config: SovereignConfig<Self::Api>) {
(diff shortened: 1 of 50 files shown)


chain-config/src/lib.rs
```diff
@@ -17,10 +21,28 @@ pub trait ChainConfigContract:
     }
 
     #[only_owner]
-    #[endpoint(updateConfig)]
-    fn update_config(&self, new_config: SovereignConfig<Self::Api>) {
+    #[endpoint(updateSovereignConfig)]
+    fn update_sovereign_config(
+        &self,
+        hash_of_hashes: ManagedBuffer,
+        new_config: SovereignConfig<Self::Api>,
```
axenteoctavian, line 28: This is not ok.
The old endpoint was fine because can to be called from sovereign-forge.

You need to create a different endpoint:
- should not be `only_owner` because is called by the bridge service
- just use `self.require_setup_complete();`
- lock_operation_hash
- `require_valid_config` should not fail the transaction because we need to generate event
- set config if valid
- remove_executed_hash
- generate `executedBridgeOp` event

axenteoctavian: Maybe refactor `require_valid_config` function to return `(bool, string)` [is_valid, err_message]
- for the old endpoint you can use `sc_panic` if not valid
- for this endpoint, we need to generate a different event with the error message

axenteoctavian (changes requested):

chain-config/src/lib.rs
```diff
@@ -1,14 +1,18 @@
...
 #[multiversx_sc::contract]
 pub trait ChainConfigContract:
-    validator_rules::ValidatorRulesModule + setup_phase::SetupPhaseModule
+    validator_rules::ValidatorRulesModule
+    + setup_phase::SetupPhaseModule
+    + execute_common::ExecuteCommonModule
+    + storage::CrossChainStorage
 {
     #[init]
     fn init(&self, config: SovereignConfig<Self::Api>) {
```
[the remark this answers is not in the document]
andreiblt1304: Well, isn't it mandatory for a Sovereign Chain to have a set configuration? If not, one improvement would be to have the same logic as in the mvx-esdt-safe contract, where if the config parameter is `None` we put the default configuration.

chain-config/src/lib.rs
```diff
@@ -1,14 +1,18 @@
...
 
 pub mod validator_rules;
 
 #[multiversx_sc::contract]
 pub trait ChainConfigContract:
-    validator_rules::ValidatorRulesModule + setup_phase::SetupPhaseModule
+    validator_rules::ValidatorRulesModule
+    + setup_phase::SetupPhaseModule
+    + execute_common::ExecuteCommonModule
+    + storage::CrossChainStorage
```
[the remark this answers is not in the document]
andreiblt1304: It was impossible to make a common cross-chain module without adding the storages as well. They work the as the common-error module.

chain-config/src/lib.rs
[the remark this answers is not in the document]
axenteoctavian: I think we can use he same logic from mvx-esdt-safe.

chain-config/src/lib.rs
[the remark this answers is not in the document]
axenteoctavian: the common-error module is just constant I think. I don't think is ok to import all the storages in all contracts. This approach seems bad and prone to errors.

chain-config/src/validator_rules.rs
```diff
@@ -25,6 +26,30 @@ pub trait ValidatorRulesModule {
...
 
+    fn is_new_config_valid(&self, config: &SovereignConfig<Self::Api>) -> Option<&str> {
+        if config.min_validators <= config.max_validators {
+            None
+        } else {
+            Some(INVALID_MIN_MAX_VALIDATOR_NUMBERS)
+        }
+    }
+
+    fn lock_operation_hash(&self, hash_of_hashes: &ManagedBuffer, hash: &ManagedBuffer) {
```
axenteoctavian, line 37: This function should be in a common module because it's used in multiple contracts

chain-config/src/validator_rules.rs
```diff
@@ -25,6 +26,30 @@ pub trait ValidatorRulesModule {
...
+        self.tx()
+            .to(self.blockchain().get_owner_address())
+            .typed(HeaderverifierProxy)
+            .lock_operation_hash(hash_of_hashes, hash)
+            .sync_call();
+    }
+
+    fn remove_executed_hash(&self, hash_of_hashes: &ManagedBuffer, op_hash: &ManagedBuffer) {
```
axenteoctavian, line 45: This function should be in a common module because it's used in multiple contracts

chain-config/src/lib.rs
```diff
@@ -17,12 +18,38 @@ pub trait ChainConfigContract:
...
+            self.failed_bridge_operation_event(
+                &hash_of_hashes,
+                &config_hash,
+                &ManagedBuffer::from(error_message),
+            );
+
+            return;
+        }
+
+        self.sovereign_config().set(new_config);
```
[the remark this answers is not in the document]
axenteoctavian: same for esdt-safe

chain-config/src/validator_rules.rs
```diff
@@ -25,6 +26,30 @@ pub trait ValidatorRulesModule {
         );
     }
 
+    fn is_new_config_valid(&self, config: &SovereignConfig<Self::Api>) -> Option<&str> {
```
axenteoctavian, line 29: It's better to combine this function with `require_validator_range`. I made a suggestion in comments above for this.

axenteoctavian (changes requested):
And remaining comments are important

chain-config/tests/chain_config_blackbox_tests.rs
```diff
@@ -70,3 +76,125 @@ fn test_complete_setup_phase() {
 
     state.common_setup.complete_chain_config_setup_phase(None);
 }
+
+/// ### TEST
+/// C-CONFIG_UPDATE_CONFIG_FAIL_004
+///
+/// ### ACTION
+/// Call 'update_chain_config()' during the setup phase
```
sergiuosvat, line 84: Wrong function name, should be `update_sovereign_config`

chain-config/tests/chain_config_blackbox_tests.rs
```diff
@@ -70,3 +76,125 @@ fn test_complete_setup_phase() {
...
+        Some(SETUP_PHASE_NOT_COMPLETED),
+        None,
+    );
+}
+
+/// ### TEST
+/// C-CONFIG_UPDATE_CONFIG_FAIL_005
+///
+/// ### ACTION
+/// Call 'update_chain_config()'  with an invalid config
```
sergiuosvat, line 116: Same here and i think the config is valid, maybe wrong test name?

chain-config/tests/chain_config_blackbox_tests.rs
```diff
@@ -70,3 +76,125 @@ fn test_complete_setup_phase() {
...
+        MultiValueEncoded::from_iter(vec![config_hash]),
+    );
+
+    state.common_setup.complete_chain_config_setup_phase(None);
+
+    state.update_sovereign_config(hash_of_hashes, new_config, None, Some("failedBridgeOp"));
+}
+
+/// ### TEST
+/// C-CONFIG_UPDATE_CONFIG_FAIL_005
```
sergiuosvat, line 153: Here the test does not fail, wrong test name and also wrong function in ACTION

mvx-esdt-safe/tests/mvx_esdt_safe_blackbox_tests.rs
```diff
@@ -2373,3 +2374,189 @@ fn test_set_token_lock_mechanism_token_from_sovereign() {
...
+/// ### ACTION
+/// Call `update_esdt_safe_config()` before registering operation
+///
+/// ### EXPECTED
+/// ERROR CURRENT_OPERATION_NOT_REGISTERED
+#[test]
+fn test_update_config_operation_not_registered() {
+    let mut state = MvxEsdtSafeTestState::new();
+    state.deploy_contract_with_roles();
+    // state.common_setup.deploy_chain_config(config)
```
sergiuosvat, line 2419: Maybe delete this line if it is not needed

chain-config/src/validator_rules.rs
andreiblt1304: The reason I removed the these functions is to remove the import of the common storage module. We have 2 choices: 
1. Import the common module that has these functions but we have to import the storage module as well
2. Create a new local function for each contract

Another research topic that we can look into later is that we implement a trait for hash locking just as the `GenearteHash` trait.

chain-config/src/validator_rules.rs
andreiblt1304: Same response as the comment above

chain-config/src/validator_rules.rs
andreiblt1304: Instead of combining both functions is easier to use the `is_config_valid` that returns an `Option` value. In the `init` function there will be a `require` that checks if the function return value `is_none`, if not `panic` with the error message

chain-config/src/lib.rs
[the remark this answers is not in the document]
andreiblt1304: This should be done in another PR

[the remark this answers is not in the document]
axenteoctavian: Ok. And please use the logic from `update_sovereign_config_during_setup_phase`

PR #361, opened 2025-05-23
Add check for returning logs without expecting them

sergiuosvat:


andreiblt1304:
I think it's better to have a `match` then so many ifs and elses.

```
match expected_log {
    Some(expected) => self.assert_expected_log(logs, expected),
    None => {
        assert!(
            logs.is_empty(),
            "Expected no logs, but got: {:?}",
            logs
        );
    }
}
```

Maybe this can also be put inside a function so we get rid of duplicated code.

fgandila (approved):

PR #371, opened 2025-05-28
Setup phases reorder

andreiblt1304:
Since the header-verifier will have a locking and removing logic for operation hashes and that these operations can only be called by the smart contracts inside the Sovereign Chain a new mapper and checks were introduced. 

Because the array of contract ID and Address has to be inside the header-verifier's storage, this contract will now be deployed last inside the sovereign-forge. 

The new order of deployment will look like this:
* Phase one: chain-config
* Phase two: either mvx-esdt-safe ori enshrine-esdt-safe
* Phase three: fee-market
* Phase four: header-verifier

### DIFF
--- a/sovereign-forge/tests/sovereign_forge_blackbox_tests.rs
+++ b/sovereign-forge/tests/sovereign_forge_blackbox_tests.rs
@@ -335,13 +335,22 @@ fn test_update_esdt_safe_config() {
 fn test_set_fee() {
     let mut state = SovereignForgeTestState::new();
     state.common_setup.deploy_sovereign_forge();
-    state.common_setup.deploy_chain_factory();
+
     state
         .common_setup
         .deploy_chain_config(OptionalValue::None, None);
+
+    state
+        .common_setup
+        .deploy_header_verifier(CHAIN_CONFIG_ADDRESS);
+
+    state.common_setup.deploy_mvx_esdt_safe(OptionalValue::None);
+
     state
         .common_setup
         .deploy_fee_market(None, ESDT_SAFE_ADDRESS);
+
+    state.common_setup.deploy_chain_factory();
     state.finish_setup();
 
     let deploy_cost = BigUint::from(100_000u32);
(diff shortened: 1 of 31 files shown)


sovereign-forge/tests/sovereign_forge_blackbox_tests.rs
```diff
@@ -955,26 +917,25 @@ fn test_deploy_phase_three() {
...
         .world
         .query()
         .to(SOVEREIGN_FORGE_SC_ADDRESS)
         .whitebox(sovereign_forge::contract_obj, |sc| {
-            let is_esdt_safe_deployed = sc.is_contract_deployed(
-                &OWNER_ADDRESS.to_managed_address(),
-                ScArrayFromUtils::ESDTSafe,
-            );
+            let is_esdt_safe_deployed =
+                sc.is_contract_deployed(&OWNER_ADDRESS.to_managed_address(), ScArray::ESDTSafe);
```
sergiuosvat, line 938: Maybe check here for fee market address, we know that esdt-safe is already deployed due to the `require_phase_two_completed` in `deploy_phase_three`

sovereign-forge/tests/sovereign_forge_blackbox_tests.rs
```diff
@@ -1088,7 +1051,7 @@ fn test_complete_setup_phase_four_not_deployed() {
     state
         .common_setup
         .deploy_fee_market(None, ESDT_SAFE_ADDRESS);
-    state.complete_setup_phase(Some(FEE_MARKET_NOT_DEPLOYED));
+    state.complete_setup_phase(Some(HEADER_VERIFIER_NOT_DEPLOYED));
 }
 
 /// ### TEST
```
sergiuosvat, line 1057: The test name is `deploy_phase_four` but the actual call to deploy phase four does not exist

sovereign-forge/tests/sovereign_forge_blackbox_tests.rs
```diff
@@ -1169,15 +1131,19 @@ fn test_deploy_phase_four_without_previous_phase() {
...
     state
         .common_setup
-        .deploy_phase_four(None, Some(ESDT_SAFE_NOT_DEPLOYED));
+        .deploy_phase_two(OptionalValue::None, None);
+    state
+        .common_setup
+        .deploy_phase_four(Some(FEE_MARKET_NOT_DEPLOYED));
 }
 
 /// ### TEST
```
sergiuosvat, line 1149: Here as well, the test description does not match the test scope

axenteoctavian (approved):

sergiuosvat (approved):

PR #376, opened 2025-06-03
Fee-Market hash locking logic

andreiblt1304:
Added hash-locking for the endpoints that will be used after the completion of the sovereign setup phase. The endpoints modified are: `distribute_fees`, `set_fee`, `remove_fee`. For the latter 2, there was a copy added to be used during the setup-phase.

Now each endpoint includes `hash_of_hashes` as parameter, the ability to lock the hash of the current operation, emit failed or confirmation events and remove the executed hash from the header-verifier SC.

### DIFF
--- a/fee-market/src/subtract_fee.rs
+++ b/fee-market/src/subtract_fee.rs
@@ -115,17 +115,18 @@ pub trait SubtractFeeModule:
 
     #[payable("*")]
     #[endpoint(subtractFee)]
-    fn subtract_fee(&self, fee_context: FeeContext<Self::Api>) -> FinalPayment<Self::Api> {
+    fn subtract_fee(
+        &self,
+        original_caller: ManagedAddress,
+        total_transfers: usize,
+        opt_gas_limit: OptionalValue<GasLimit>,
+    ) -> FinalPayment<Self::Api> {
         self.require_caller_esdt_safe();
 
         let caller = self.blockchain().get_caller();
         let payment = self.call_value().single_esdt().clone();
 
-        if !self.is_fee_enabled()
-            || self
-                .users_whitelist()
-                .contains(&fee_context.original_caller)
-        {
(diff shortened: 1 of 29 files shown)


sergiuosvat (approved):

fee-market/src/subtract_fee.rs
```diff
@@ -18,6 +19,8 @@ pub trait SubtractFeeModule:
     + crate::fee_common::CommonFeeModule
     + crate::price_aggregator::PriceAggregatorModule
     + utils::UtilsModule
+    + setup_phase::SetupPhaseModule
+    + events::EventsModule
 {
     #[only_owner]
```
axenteoctavian, line 25: Not needed

fee-market/tests/fee_market_blackbox_test.rs
```diff
@@ -40,25 +55,638 @@ fn test_set_fee_wrong_params() {
...
+        .deploy_header_verifier(vec![ScArray::FeeMarket, ScArray::ChainConfig]);
 
-    state.set_fee(SECOND_TEST_TOKEN, "Fixed", Some(INVALID_FEE));
+    let fee = FeeStruct {
+        base_token: FIRST_TEST_TOKEN.to_token_identifier(),
+        fee_type: FeeType::None,
+    };
 
-    state.set_fee(FIRST_TEST_TOKEN, "AnyTokenWrong", Some(INVALID_TOKEN_ID));
+    state.set_fee(&ManagedBuffer::new(), &fee, Some(CALLER_NOT_OWNER), None);
```
axenteoctavian, line 100: This should be `SETUP_PHASE_NOT_COMPLETED` (after you remove only_owner for the function)

axenteoctavian: Applicable for other tests below

PR #382, opened 2025-06-06
Sovereign cleanup and coverage

andreiblt1304:


sergiuosvat (approved):

sovereign-forge/tests/sovereign_forge_blackbox_tests.rs
axenteoctavian, line 1: I see in all tests, you deploy the contracts, then you also call some deploy phases. That's not ok.
The clean way would be to deploy sovereign-forge and chain-factory inside `SovereignForgeTestState::new();`, then call deploy phases

PR #385, opened 2025-06-11
Emit failed event instead of panic when hash generation fails

andreiblt1304:


CostinCarabas (approved):

chain-config/src/lib.rs
```diff
@@ -58,6 +67,9 @@ pub trait ChainConfigContract:
                 &config_hash,
                 &ManagedBuffer::from(error_message),
             );
+
+            self.remove_executed_hash(&hash_of_hashes, &config_hash);
+            return;
```
axenteoctavian, line 72: This is not needed, you call it 2 times

PR #387, opened 2025-06-12
Validator registration endpoints

andreiblt1304:


chain-config/src/validator_rules.rs
```diff
@@ -21,10 +24,91 @@ pub trait ValidatorRulesModule {
...
     }
 
+    #[endpoint(register)]
+    fn register(&self, new_validator: ValidatorInfo<Self::Api>) {
+        self.require_setup_complete();
+        self.require_validator_not_registered(&new_validator.bls_key);
+
+        let max_number_of_validators = self.sovereign_config().get().max_validators;
+        let last_bls_key_id = self.last_bls_key_id().get();
+        let current_bls_key_id = &last_bls_key_id + 1u32;
```
axenteoctavian, line 34: We should start from zero

chain-config/src/validator_rules.rs
```diff
@@ -21,10 +24,91 @@ pub trait ValidatorRulesModule {
...
+    }
+
+    #[endpoint(unregister)]
+    fn unregister(&self, validator_info: ValidatorInfo<Self::Api>) {
+        self.require_setup_complete();
+        self.require_validator_registered(&validator_info.bls_key);
+
+        let min_number_of_validators = self.sovereign_config().get().min_validators;
+        let last_bls_key_id = self.last_bls_key_id().get();
+        let current_bls_key_id = &last_bls_key_id - 1u32;
```
axenteoctavian, line 62: We should not do this. The ID si always incrementing

chain-config/src/validator_rules.rs
```diff
@@ -21,10 +24,91 @@ pub trait ValidatorRulesModule {
...
+
+        require!(
+            current_bls_key_id >= min_number_of_validators,
```
axenteoctavian, line 65: This is not really ok. Validators should be able to unregister.

chain-config/tests/chain_config_blackbox_tests.rs
```diff
@@ -289,3 +294,258 @@ fn test_update_config() {
...
+///
+/// ### EXPECTED
+/// Error VALIDATOR_RANGE_EXCEEDED
+#[test]
+fn test_register_validator_range_exceeded_too_many_validators() {
+    let mut state = ChainConfigTestState::new();
+
+    state
+        .common_setup
+        .deploy_chain_config(OptionalValue::None, None);
```
axenteoctavian, line 338: Please use here `default_config` to be more clear that max number of validators is 1

chain-config/tests/chain_config_blackbox_tests.rs
```diff
@@ -289,3 +294,258 @@ fn test_update_config() {
...
+/// ### TEST
+/// C-CONFIG_REGISTER_VALIDATOR_OK
+///
+/// ### ACTION
+/// Call 'register()' with valid validator
+///
+/// ### EXPECTED
+/// Validator is registered successfully
+#[test]
+fn test_register_validator() {
```
axenteoctavian, line 402: Register multiple validators in this test

common/structs/src/lib.rs
```diff
@@ -29,3 +29,12 @@ pub struct IssueEsdtArgs<M: ManagedTypeApi> {
...
     pub num_decimals: usize,
 }
+
+#[type_abi]
+#[derive(TopEncode, TopDecode, NestedEncode)]
+pub struct ValidatorInfo<M: ManagedTypeApi> {
+    pub address: ManagedAddress<M>,
+    pub bls_key: ManagedBuffer<M>,
+    pub egld_stake: BigUint<M>,
+    pub token_stake: EsdtTokenData<M>,
```
axenteoctavian, line 39: I don't think we'll need this. We'll see later

common/events/src/lib.rs
```diff
@@ -36,4 +36,23 @@ pub trait EventsModule {
         #[indexed] hash: &ManagedBuffer,
         error_message: &ManagedBuffer,
     );
+
+    // address, blsKeys, eGLDStake, tokenStake
+    #[event("register")]
+    fn register_event(
```
axenteoctavian, line 42: We need to send the `id` in both events

axenteoctavian (changes requested):
TODOs:
- complete_setup_phase should validate the config
-  we might also need some `pause` module to temporary stop the registrations

chain-config/src/validator_rules.rs
```diff
@@ -21,10 +24,87 @@ pub trait ValidatorRulesModule {
...
+        let last_bls_key_id_mapper = self.last_bls_key_id();
+        let current_bls_key_id = &last_bls_key_id_mapper.get() + 1u32;
+
+        require!(
+            current_bls_key_id <= max_number_of_validators,
+            VALIDATOR_RANGE_EXCEEDED
+        );
+
+        self.last_bls_key_id()
+            .update(|id| *id += BigUint::from(1u32));
```
axenteoctavian, line 42: You can update with `current_bls_key_id`

chain-config/src/validator_rules.rs
axenteoctavian: There is no need for an event. The min number of validators will be checked in other places.

axenteoctavian (commented):
Please create tasks for:
- `complete_setup_phase` should validate the config
- also I'm thinking about this flow (not sure if this is right, we need to discuss) -> there will be more tasks
Registrations flow:
- sovereign owner will complete_setup_phase in sovereign-forge
- validators will come and register in chain-config
-> we didn't consider `opt_additional_stake_required` from config
- sovereign owner will disable the registrations
-> he should be able to disable only once (if min validators number is met)
-> these are the genesis validators and need to be added in header-verifier for epoch 0
- the genesis validators will be extracted from contract (read from storage/call some endpoint) to have them for node config
- the sovereign chain will be deployed
- then an outgoing operations will enable the register/unregister endpoints

PR #390, opened 2025-06-19
Added validator logic to header-verifier

andreiblt1304:


interactor/tests/sovereign_forge_tests.rs
andreiblt1304, line 1: Those test were ignored to reduce conflicts with the [Cross-Shard interactor](https://github.com/multiversx/mx-sovereign-sc/pull/402) PR.

header-verifier/src/lib.rs
```diff
@@ -201,19 +219,92 @@ pub trait Headerverifier: events::EventsModule + setup_phase::SetupPhaseModule {
...
+
+        let bitmap_byte_array = &padded_bitmap_byte_array[..bls_keys_length];
+
+        let mut approving_validators_bls_keys: ManagedVec<Self::Api, ManagedBuffer> =
+            ManagedVec::new();
+
+        for (index, has_signed) in bitmap_byte_array.iter().enumerate() {
+            if *has_signed == 1u8 {
+                approving_validators_bls_keys.push(
+                    self.bls_keys_map(self.get_chain_config_address())
```
axenteoctavian, line 268: You need to get BLS key from `self.bls_pub_keys(epoch)`, have `epoch` as argument in this function

header-verifier/src/lib.rs
```diff
@@ -201,19 +219,92 @@ pub trait Headerverifier: events::EventsModule + setup_phase::SetupPhaseModule {
...
-        pub_keys_count > minimum_signatures
+        bls_keys
+    }
+
+    fn get_approving_validators(
+        &self,
+        bls_keys_bitmap: &ManagedBuffer,
+        bls_keys_length: usize,
+    ) -> ManagedVec<ManagedBuffer> {
+        let mut padded_bitmap_byte_array = [0u8; 1024];
```
axenteoctavian, line 257: `1024` should be the value of `SovereignConfig.max_validators`

header-verifier/src/lib.rs
```diff
@@ -201,19 +219,92 @@ pub trait Headerverifier: events::EventsModule + setup_phase::SetupPhaseModule {
...
+                );
+            }
+        }
+
+        let minimum_signatures = 2 * bls_keys_length / 3 + 1;
+
+        require!(
+            approving_validators_bls_keys.len() >= minimum_signatures,
+            MIN_NUMBER_OF_SIGNATURE_NOT_MET
+        );
```
axenteoctavian, line 280: This is not needed here because is always true.
If less than 2/3+1 validators sign, the block is not proposed and the outgoing operations are not sent.

header-verifier/src/lib.rs
```diff
@@ -76,28 +81,51 @@ pub trait Headerverifier: events::EventsModule + setup_phase::SetupPhaseModule {
...
         operation_hash: ManagedBuffer,
-        _pub_keys_bitmap: ManagedBuffer,
-        _epoch: ManagedBuffer,
-        _pub_keys_id: MultiValueEncoded<ManagedBuffer>,
+        pub_keys_bitmap: ManagedBuffer,
+        epoch: u64,
+        pub_keys_id: MultiValueEncoded<BigUint<Self::Api>>,
     ) {
+        self.require_setup_complete();
+        require!(
```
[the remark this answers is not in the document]
andreiblt1304: I will add the events after the merging of events, I will put a TODO for now

header-verifier/src/lib.rs
andreiblt1304: You can't put a variable for this byte array length, it has to be a constant value

header-verifier/tests/header_verifier_blackbox_tests.rs
```diff
@@ -459,22 +523,69 @@ fn test_lock_operation_hash_already_locked() {
...
+
+    let bitmap = ManagedBuffer::new_from_bytes(&[1]);
+
     state.change_validator_set(
         &ManagedBuffer::new(),
         &hash_of_hashes,
         &operation_hash,
+        1,
+        &bitmap,
+        validator_set,
```
axenteoctavian, line 588: Please validate the bls mapper for epoch 1 after this

header-verifier/tests/header_verifier_blackbox_tests.rs
axenteoctavian, line 1: Please add one more test where you change_validators_set in a loop (10 times or something) and validate that bls mapper is cleaning for old epochs

PR #392, opened 2025-06-23
Validator genesis registration

andreiblt1304:
During the setup-phase of a Sovereign Chain there will a so called "genesis phase" that will allow the first whitelisted validators to register into the Chain-Config contract to validate the newly created Sovereign-Chain.

This PR includes the genesis registration logic for the Chain-Config SC. It added a new endpoint `complete_genesis_phase` and a new storage flag `genesis_phase`. The flag is used inside the `register` endpoint.


chain-config/src/validator_rules.rs
```diff
@@ -104,6 +148,9 @@ pub trait ValidatorRulesModule: setup_phase::SetupPhaseModule + events::EventsMo
     #[storage_mapper("lastBlsKeyId")]
     fn last_bls_key_id(&self) -> SingleValueMapper<BigUint<Self::Api>>;
 
+    #[storage_mapper("genesisPhase")]
+    fn genesis_phase(&self) -> SingleValueMapper<bool>;
```
axenteoctavian, line 152: I don't think this naming is correct. It should be like `registration_status` or something similar (enabled/disable)

chain-config/src/validator_rules.rs
```diff
@@ -17,18 +18,33 @@ pub struct TokenIdAmountPair<M: ManagedTypeApi> {
...
         if config.min_validators <= config.max_validators {
             None
         } else {
             Some(INVALID_MIN_MAX_VALIDATOR_NUMBERS)
         }
     }
 
+    #[payable]
     #[endpoint(register)]
     fn register(&self, new_validator: ValidatorInfo<Self::Api>) {
```
axenteoctavian, line 36: I believe the validations should be:
- validate if enabled/disabled
- validate additional stake 
- require_validator_not_registered
- the rest is fine

axenteoctavian: I think you can make another PR for `unregister` where you check enable/disabled and return the additional stake

axenteoctavian: Also the endpoint for enabling/disabling registrations which is done with 2 steps - register + update(or change_status or something similar)

axenteoctavian: Also the endpoint to read the genesis validators

chain-config/src/lib.rs
```diff
@@ -78,6 +79,20 @@ pub trait ChainConfigContract:
         self.execute_bridge_operation_event(&hash_of_hashes, &config_hash);
     }
 
+    #[only_owner]
+    #[endpoint(completeGenesis)]
+    fn complete_genesis(&self) {
```
axenteoctavian, line 84: I'm thinking we can use the `complete_setup_phase` function instead of this

chain-config/src/validator_rules.rs
```diff
@@ -86,6 +102,34 @@ pub trait ValidatorRulesModule: setup_phase::SetupPhaseModule + events::EventsMo
...
+    }
+
+    fn require_caller_whitelisted(&self) {
+        if let Some(additional_stake) = &self.sovereign_config().get().opt_additional_stake_required
+        {
+            let call_value = self.call_value().all_esdt_transfers();
+
+            for stake in additional_stake {
+                let matched = call_value.iter().any(|paid| {
+                    paid.token_identifier == stake.token_id && paid.amount >= stake.amount
```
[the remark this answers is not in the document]
andreiblt1304: I'll mark this as a TODO

chain-config/src/validator_rules.rs
axenteoctavian, line 36:
```suggestion
    fn register(&self, bls_key: ManagedBuffer) {
```
You can create `ValidatorInfo` from caller + bls_key + payment

chain-config/src/lib.rs
```diff
@@ -78,13 +81,66 @@ pub trait ChainConfigContract:
...
+            self.remove_executed_hash(&hash_of_hashes, &status_hash);
+
+            return;
+        }
+
+        self.lock_operation_hash(&status_hash, &hash_of_hashes);
+
+        let registration_status_mapper = self.registration_status();
+
+        if registration_status == registration_status_mapper.get() {
```
axenteoctavian, line 111: I don't think this check is really necessary, it's just more code and more tests not needed

chain-config/src/lib.rs
```diff
@@ -78,13 +81,66 @@ pub trait ChainConfigContract:
...
+
+        registration_status_mapper.set(registration_status);
+
+        self.remove_executed_hash(&hash_of_hashes, &status_hash);
+        self.registration_status_update_event();
+    }
+
+    // #[only_owner]
+    // #[endpoint(freezeRegistration)]
+    // fn freeze_registration
```
axenteoctavian, line 131: Is this still needed?

chain-config/src/validator_rules.rs
```diff
@@ -40,22 +52,22 @@ pub trait ValidatorRulesModule: setup_phase::SetupPhaseModule + events::EventsMo
...
-            &new_validator.egld_stake,
-            &new_validator.token_stake,
+            &self.blockchain().get_caller(),
+            &new_bls_key,
+            &egld_stake,
+            &additional_stake,
         );
     }
 
+    // TODO: add storage for registered stake in order to return it upon unregistering
```
axenteoctavian, line 68: We said we can save the registered stake into ValidatorInfo and set it in the MapMapper

common/events/src/lib.rs
```diff
@@ -54,6 +56,12 @@ pub trait EventsModule {
...
         #[indexed] egld_stake: &BigUint,
-        #[indexed] token_stake: &EsdtTokenData<Self::Api>,
+        #[indexed] token_stake: &ManagedVec<TokenStake<Self::Api>>,
     );
+
+    #[event("completeGenesisPhase")]
+    fn complete_genesis_event(&self);
+
+    #[event("registrationStatusUpdate")]
+    fn registration_status_update_event(&self);
```
axenteoctavian, line 66: Please put the status as string in this event

chain-config/src/validator_rules.rs
```diff
@@ -40,22 +51,22 @@ pub trait ValidatorRulesModule: setup_phase::SetupPhaseModule + events::EventsMo
 
         self.last_bls_key_id().set(current_bls_key_id.clone());
         self.bls_keys_map()
-            .insert(current_bls_key_id.clone(), new_validator.bls_key.clone());
-        self.bls_key_to_id_mapper(&new_validator.bls_key)
+            .insert(current_bls_key_id.clone(), new_bls_key.clone());
```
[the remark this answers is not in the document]
andreiblt1304: Payment saving will be done in a different PR

chain-config/src/validator_rules.rs
```diff
@@ -86,6 +96,56 @@ pub trait ValidatorRulesModule: setup_phase::SetupPhaseModule + events::EventsMo
...
+            require!(matched, INVALID_ADDITIONAL_STAKE);
+        }
+
+        (
+            egld_amount.clone(),
+            sovereign_config.opt_additional_stake_required,
+        )
+    }
+
+    fn require_registration_disabled(&self) {
```
axenteoctavian, line 145:
```suggestion
    fn require_registration_enabled(&self) {
```

common/structs/src/lib.rs
```diff
@@ -36,5 +36,30 @@ pub struct ValidatorInfo<M: ManagedTypeApi> {
     pub address: ManagedAddress<M>,
     pub bls_key: ManagedBuffer<M>,
     pub egld_stake: BigUint<M>,
-    pub token_stake: EsdtTokenData<M>,
+    pub token_stake: ManagedVec<M, TokenStake<M>>,
+}
+
+#[type_abi]
+#[derive(TopEncode, TopDecode, NestedEncode, NestedDecode, ManagedVecItem)]
+pub struct TokenStake<M: ManagedTypeApi> {
```
axenteoctavian, line 44: Why do you need `TokenStake`? You can just use `EsdtTokenPayment`

common/structs/src/configs.rs
```diff
@@ -13,7 +13,7 @@ pub struct SovereignConfig<M: ManagedTypeApi> {
     pub min_validators: u64,
     pub max_validators: u64,
     pub min_stake: BigUint<M>,
-    pub opt_additional_stake_required: Option<ManagedVec<M, StakeArgs<M>>>,
+    pub opt_additional_stake_required: Option<ManagedVec<M, EsdtTokenPayment<M>>>,
```
axenteoctavian, line 16: It was ok with `StakeArgs` here because only id + amount should be defined here.

chain-config/src/validator_rules.rs
```diff
@@ -86,6 +96,56 @@ pub trait ValidatorRulesModule: setup_phase::SetupPhaseModule + events::EventsMo
...
+                    .iter()
+                    .any(|c| c.token_identifier == s.token_identifier && c.amount >= s.amount)
+            });
+
+            require!(matched, INVALID_ADDITIONAL_STAKE);
+        }
+
+        (
+            egld_amount.clone(),
+            sovereign_config.opt_additional_stake_required,
```
axenteoctavian, line 141: Here you can return the esdt transfers from call_value (without EGLD-000000)

chain-config/src/validator_rules.rs
```diff
@@ -86,6 +96,56 @@ pub trait ValidatorRulesModule: setup_phase::SetupPhaseModule + events::EventsMo
...
+    ) {
+        let sovereign_config = self.sovereign_config().get();
+
+        let call_value = self.call_value().all_transfers().clone_value();
+        let mut egld_amount = BigUint::zero();
+
+        if sovereign_config.min_stake > 0 {
+            let egld_payment = call_value
+                .iter()
+                .find(|p| p.token_identifier.is_egld() && p.amount >= sovereign_config.min_stake);
```
axenteoctavian, line 122: Still not resolved, should be `==`

chain-config/src/lib.rs
```diff
@@ -78,13 +82,51 @@ pub trait ChainConfigContract:
...
+        }
+
+        let registration_status_mapper = self.registration_status();
+
+        registration_status_mapper.set(registration_status);
+
+        self.remove_executed_hash(&hash_of_hashes, &status_hash);
+        self.registration_status_update_event(&ManagedBuffer::new_from_bytes(&[
+            registration_status,
+        ]));
```
axenteoctavian, line 117: Can you set the buffer like string 'enabled' or 'disabled' instead of '0' or '1'?

chain-config/src/validator_rules.rs
```diff
@@ -86,6 +96,61 @@ pub trait ValidatorRulesModule: setup_phase::SetupPhaseModule + events::EventsMo
...
+    fn require_validator_set_valid(&self, validator_len: usize) {
+        let config = self.sovereign_config().get();
+
+        require!(
+            validator_len as u64 >= config.min_validators,
+            NOT_ENOUGH_VALIDATORS
+        );
+    }
+
+    fn validate_stake(
```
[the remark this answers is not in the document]
axenteoctavian: Also please define the constant
```
pub const ADDITIONAL_STAKE_NOT_REQUIRED: &str = "Additional stake was provided but is not required";
```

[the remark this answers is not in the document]
axenteoctavian: After this, some tests should fail because you set min_stake=0 in sovereign_config but you provide egld_payment in register
e.g. `test_register_validator_is_whitelisted`

axenteoctavian (changes requested):

PR #393, opened 2025-06-27
Sovereign-forge async deploy calls

andreiblt1304:
Since there will be only one instance for the sovereign-forge across the network, the deploy calls should be async.


sergiuosvat (approved):

alyn509 (approved):

sovereign-forge/src/common/sc_deploy.rs
```diff
@@ -1,67 +1,118 @@
...
         config: OptionalValue<SovereignConfig<Self::Api>>,
-    ) -> ManagedAddress {
+    ) {
         self.tx()
             .to(self.get_chain_factory_address())
             .typed(ChainFactoryContractProxy)
             .deploy_sovereign_chain_config_contract(config)
-            .returns(ReturnsResult)
-            .sync_call()
+            .gas(self.blockchain().get_gas_left() / 2)
```
axenteoctavian, line 26: Can you use like this?
```suggestion
            .gas(self.blockchain().get_gas_left())
```

or use a default constant value

sovereign-forge/src/common/sc_deploy.rs
```diff
@@ -1,67 +1,118 @@
...
+            .callback(
+                self.callbacks()
+                    .register_deployed_contract(chain_id, ScArray::ChainConfig),
+            )
+            .gas_for_callback(15_000_000)
```
axenteoctavian, line 31: We need some test to make this gas as efficient as possible

sovereign-forge/src/common/sc_deploy.rs
```diff
@@ -1,67 +1,118 @@
...
+    ) {
+        match result {
+            ManagedAsyncCallResult::Ok(sc_address) => {
+                let new_contract_info = ContractInfo::new(sc_id, sc_address);
+
+                self.sovereign_deployed_contracts(chain_id)
+                    .insert(new_contract_info);
+            }
+            ManagedAsyncCallResult::Err(_) => {
+                sc_panic!("");
```
axenteoctavian, line 114: Please sc_panic with returned error message, not empty

sovereign-forge/src/common/sc_deploy.rs
axenteoctavian: Then probably is best to have a constant value for deploy

PR #430, opened 2025-08-21
Fee common module

andreiblt1304:


sergiuosvat (approved):

common/fee-common/src/endpoints.rs
```diff
@@ -0,0 +1,79 @@
...
+#[multiversx_sc::module]
+pub trait FeeCommonEndpointsModule:
+    crate::helpers::FeeCommonHelpersModule
+    + crate::storage::FeeCommonStorageModule
+    + utils::UtilsModule
+    + custom_events::CustomEventsModule
+{
+    fn distribute_fees_common_function(
+        &self,
+        hash_of_hashes: &ManagedBuffer,
```
axenteoctavian, line 15: `hash_of_hashes` should be in common because for sov-esdt-safe is not needed

fee-market/src/fee_operations.rs
```diff
@@ -0,0 +1,38 @@
...
+        hash_of_hashes: ManagedBuffer,
+        address_percentage_pairs: MultiValueEncoded<MultiValue2<ManagedAddress, usize>>,
+    ) {
+        self.require_setup_complete();
+        self.distribute_fees_common_function(&hash_of_hashes, address_percentage_pairs);
+    }
+
+    #[payable("*")]
+    #[endpoint(subtractFee)]
+    fn subtract_fee(
```
axenteoctavian, line 30: This endpoint maybe can be directly in common module

fee-market/src/fee_whitelist.rs
axenteoctavian, line 1: This should also be in common

axenteoctavian (approved):

PR #443, opened 2025-08-27
add-token-register-fix

sergiuosvat:
This PR contains changes made to the `register_token` endpoint in `mvx_esdt_safe` and a fix in log assertion


mvx-esdt-safe/src/register_token.rs
```diff
@@ -38,38 +44,40 @@ pub trait RegisterTokenModule:
             return;
         }
 
+        self.require_setup_complete_with_event(&hash_of_hashes, &token_hash);
```
andreiblt1304, line 47: please add the return logic outside this function

PR #444, opened 2025-08-28
Token registration fixes

andreiblt1304:


sergiuosvat (approved):

mvx-esdt-safe/src/register_token.rs
```diff
@@ -13,32 +20,71 @@ pub trait RegisterTokenModule:
     + cross_chain::execute_common::ExecuteCommonModule
     + custom_events::CustomEventsModule
     + multiversx_sc_modules::pause::PauseModule
+    + setup_phase::SetupPhaseModule
 {
     #[payable("EGLD")]
```
axenteoctavian, line 25: This should be deleted. The 0.05 EGLD are taken from balance

mvx-esdt-safe/src/register_token.rs
```diff
@@ -13,32 +20,71 @@ pub trait RegisterTokenModule:
...
+            self.complete_operation(
+                &hash_of_hashes,
+                &token_hash,
+                Some(SETUP_PHASE_NOT_COMPLETED.into()),
+            );
+            return;
+        }
+
+        self.lock_operation_hash_wrapper(&hash_of_hashes, &token_hash);
+
```
axenteoctavian, line 58: I see an edge case here if the sovereign is malicious. 
We need to make sure here that contract has at least `ISSUE_COST[0.05] EGLD` balance, otherwise just complete_operation with error -> NO deposit back

mvx-esdt-safe/src/register_token.rs
```diff
@@ -13,32 +20,71 @@ pub trait RegisterTokenModule:
...
+        if self.is_sov_token_id_registered(&token_to_register.token_id) {
+            self.complete_operation(
+                &hash_of_hashes,
+                &token_hash,
+                Some(TOKEN_ALREADY_REGISTERED.into()),
+            );
+
+            let tokens = self.create_event_payment_tuple(&token_to_register);
```
axenteoctavian, line 66: You need to create event with `ISSUE_COST[0.05] EGLD-000000`, not from token_to_register

mvx-esdt-safe/src/register_token.rs
```diff
@@ -13,32 +20,71 @@ pub trait RegisterTokenModule:
...
+                &token_to_register.data.op_sender.clone(),
+                &tokens,
+                token_to_register.data.clone(),
+            );
+
+            return;
+        }
+
+        if !self.has_sov_prefix(&token_to_register.token_id, &self.sov_token_prefix().get()) {
+            self.complete_operation(&hash_of_hashes, &token_hash, Some(INVALID_PREFIX.into()));
```
axenteoctavian, line 78: `INVALID_PREFIX` error message is wrong. Should be something like `provided sovereign token identifier has invalid prefix`

mvx-esdt-safe/src/register_token.rs
```diff
@@ -13,32 +20,71 @@ pub trait RegisterTokenModule:
...
+
+            return;
+        }
+
+        match token_to_register.token_type {
+            EsdtTokenType::Invalid => {
+                self.complete_operation(&hash_of_hashes, &token_hash, Some(INVALID_TYPE.into()))
+            }
+            _ => self.handle_token_issue(token_to_register, hash_of_hashes, token_hash),
         }
```
axenteoctavian, line 88: Just use handle_token_issue because this will return error event if some invalid args were provided
```suggestion
        self.handle_token_issue(token_to_register, hash_of_hashes, token_hash)
```

mvx-esdt-safe/src/register_token.rs
```diff
@@ -13,32 +20,71 @@ pub trait RegisterTokenModule:
...
     #[endpoint(registerToken)]
     fn register_token(
         &self,
-        sov_token_id: TokenIdentifier,
-        token_type: EsdtTokenType,
-        token_display_name: ManagedBuffer,
-        token_ticker: ManagedBuffer,
-        num_decimals: usize,
+        hash_of_hashes: ManagedBuffer,
+        token_to_register: SovTokenProperties<Self::Api>,
```
axenteoctavian, line 30: I think we could rename `SovTokenProperties` to `RegisterTokenOperation` 🙈

mvx-esdt-safe/src/register_token.rs
```diff
@@ -66,34 +112,57 @@ pub trait RegisterTokenModule:
...
             ManagedAsyncCallResult::Ok(mvx_token_id) => {
-                self.set_corresponding_token_ids(sov_token_id, &mvx_token_id);
+                self.set_corresponding_token_ids(&token_to_register.token_id, &mvx_token_id);
+                self.complete_operation(&hash_of_hashes, &token_hash, None);
             }
             ManagedAsyncCallResult::Err(error) => {
-                sc_panic!("There was an error at issuing token: '{}'", error.err_msg);
+                let tokens = self.create_event_payment_tuple(token_to_register);
+
+                self.deposit_event(
```
axenteoctavian, line 160: Same comment here. Deposit back `ISSUE_COST[0.05] EGLD-000000`

mvx-esdt-safe/src/register_token.rs
```diff
@@ -127,6 +196,12 @@ pub trait RegisterTokenModule:
             .set(sov_token_id);
     }
 
+    fn is_token_registered(&self, token_id: &TokenIdentifier, token_nonce: u64) -> bool {
```
axenteoctavian, line 199: This function is not used

axenteoctavian (changes requested):
Will check the tests after the fixes

axenteoctavian (approved):

PR #450, opened 2025-09-02
Modified header-verifier to emit logs

andreiblt1304:
Since the Header-Verifier smart contract will interactor with outgoing operations from the Sovereign Chain to the MultiversX mainchain, it is needed for the endpoints to either emit logs or error messages to not fail transactions.

This PR implements this logic, the endpoints right now will behave as this:
* `register_bridge_operations` will panic
* `change_validator_set` will emit bridge operation event + optional error message
* `lock/remove operation hash` will return an `OptionalValue` depicting the error message (if there was any)

### DIFF
--- a/header-verifier/src/header_utils.rs
+++ b/header-verifier/src/header_utils.rs
@@ -1,5 +1,5 @@
 use error_messages::{
-    BLS_KEY_NOT_REGISTERED, CALLER_NOT_FROM_CURRENT_SOVEREIGN, CHAIN_CONFIG_NOT_DEPLOYED,
+    BLS_KEY_NOT_REGISTERED, CHAIN_CONFIG_NOT_DEPLOYED, HASH_OF_HASHES_DOES_NOT_MATCH,
 };
 use structs::forge::ScArray;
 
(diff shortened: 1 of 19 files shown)


header-verifier/src/header_utils.rs
```diff
@@ -107,15 +111,14 @@ pub trait HeaderVerifierUtilsModule:
         //     bridge_operations_hash,
         //     signature,
         // );
+
+        None
     }
 
-    fn require_caller_is_from_current_sovereign(&self) {
+    fn is_caller_is_from_current_sovereign(&self) -> bool {
```
sergiuosvat, line 118: extra `is` in function name

header-verifier/tests/header_verifier_blackbox_tests.rs
```diff
@@ -588,7 +589,7 @@ fn test_change_validator_set() {
 /// H-VERIFIER_CHANGE_VALIDATORS_FAIL
 ///
 /// ### ACTION
-/// Call 'change_validators_set()' after registering the operation
+/// Call 'change_validator_set()' before registering the operation
```
axenteoctavian, line 592: I see this test is kind of bad...
We need to do `change_validator_set` 2 times and get `OUTGOING_TX_HASH_ALREADY_REGISTERED`,
not `register_operations` and `change_validator_set`.

axenteoctavian (commented):
We will really need blackbox tests to covert every single error from all endpoints, in a different PR

PR #479, opened 2025-09-27
Register native token by owner

andreiblt1304:


mvx-esdt-safe/src/lib.rs
axenteoctavian, line 1: I think we can remove all admins on complete setup phase

common/error-messages/src/lib.rs
```diff
@@ -158,5 +158,6 @@ pub const REGISTRATIONS_DISABLED_GENESIS_PHASE: &str =
 pub const VALIDATOR_ID_NOT_REGISTERED: &str = "Provided validator id is not registered";
 pub const INVALID_VALIDATOR_DATA: &str = "Invalid validator data has been provided";
 pub const ISSUE_COST_NOT_COVERED: &str = "Native token issue cost is not covered";
+pub const ADMIN_CAN_NOT_BE_SC: &str = "The admin of this SC can't be a SC address";
```
axenteoctavian, line 161: This is not used

axenteoctavian (approved):

PR #488, opened 2025-10-06
small fixes in the interactor

sergiuosvat:


common/common-interactor/src/interactor_common_state.rs
```diff
@@ -314,13 +293,20 @@ impl CommonState {
...
+        self.operation_nonce
+            .get(contract_address)
+            .copied()
+            .unwrap_or(0)
     }
 
-    pub fn increment_operation_nonce(&mut self, shard: u32) -> u64 {
-        self.operation_nonce[shard as usize] += 1;
-        self.operation_nonce[shard as usize]
+    pub fn increment_operation_nonce(&mut self, contract_address: &str) -> u64 {
```
axenteoctavian, line 303: Do we really need to return the incremented nonce here?

sergiuosvat: No, i'll remove it

